PR #21, opened 2025-11-10
Chore/perf nodes immer

wass08:


tahacagrimen:
Completed via PR #678. Multi-page PDF generation for saved views has been implemented.

PR #65, opened 2025-12-25
Fix edit click

frankagathos:


frankagathos:
handleNodeSelect already sets control to "edit" for "site" node type : https://github.com/pascalorg/editor/blob/main/hooks/use-editor.tsx#L788-L800

PR #68, opened 2025-12-26
Fix/single wall rotation

frankagathos:
Fixes single wall rotation behaviour to be similar to other items


wass08 (approved):
Tested & work well, thank you!

PR #69, opened 2025-12-26
Feat/make side bar resizable

frankagathos:
Adds ability to resize the side bar from the right side

### DIFF
--- a/packages/editor/app/globals.css
+++ b/packages/editor/app/globals.css
@@ -131,10 +131,11 @@
 }
 
 /* Nuclear option: disable ALL transitions and animations with maximum specificity */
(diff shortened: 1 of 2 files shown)


wass08 (changes requested):
Thank you for the PR. The flow is pretty slow:
- Use a ref to directly re-size the div without re-rendering all nodes
- The sidebar size could be saved in the editor store so it's saved accross different loads of the app

frankagathos:
@wass08 The PR was still WIP and wasn’t ready for review due to CSS transition issues during resizing.

That said, since you reviewed it, I’ve applied your suggested changes and attempted to resolve the transition issues using CSS.

PR #166, opened 2026-03-24
Add drawing measurements and measure tool

phinehas2020:


wynzn2009 (approved):

Aymericr:
Thank you for the early measurement-tool contribution. Production measurements have since shipped in #505 as persistent semantic scene nodes with the current tool/snapping architecture and dedicated tests. This branch now conflicts with main and represents the superseded pre-registry approach, so I’m closing it rather than asking for a large rebase.

Issue #147, opened 2026-03-22
Add CI workflow for lint + typecheck on push/PR


jeremylongshore:
I noticed that `biome check` and `check-types` (tsc) scripts exist at the package level, but `release.yml` is the only GitHub Actions workflow — there's no quality gate running on push or PR.

A lightweight CI workflow that runs these existing checks would catch type errors and lint issues before they hit main. Something like:

```yaml
name: CI
on:
  push: { branches: [main] }
  pull_request: { branches: [main] }
jobs:
  quality:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: oven-sh/setup-bun@v2
      - run: bun install --frozen-lockfile
      - run: bun run check
      - run: bun run check-types
```

No new tools, no new opinions — just automating what already exists.

I'd be happy to submit a PR if this would be useful.

jeremylongshore:
Hey @anton-pascal — just flagging that I'm the author of this issue and the proposed solution (including the exact workflow YAML). Since then, three separate PRs have been submitted implementing this:

- #176 by @nnhhoang (Mar 26)
- #192 by @hellozzm (Mar 28)
- #208 by @claygeo (Apr 2)

All three are implementing the spec I wrote here. @claygeo's PR #208 is the most polished — they correctly identified that `bun run lint` should be used instead of `bun run check` to avoid the 191 pre-existing format violations, which is a smart call.

Happy for any of these to land (the point is getting CI in place), but wanted to make sure the provenance is visible. Let me know if you'd prefer I submit my own PR instead.

anton-pascal:
Thanks for the heads-up @jeremylongshore — appreciate the provenance call-out.

For visibility: @Aymericr shipped #369 about an hour ago (https://github.com/pascalorg/editor/pull/369) which added the workflow and closed this issue. It's a slightly different shape from the three external PRs (28-line workflow + a few drive-by typecheck fixes that were blocking the gate), but the spirit matches what you laid out here.

PRs #192 and #208 are now stale dupes — leaving them open for their authors to decide on, but happy to close them out if that's helpful.

PR #176, opened 2026-03-26
ci: add lint and typecheck workflow on push/PR

### LINKED ISSUE #148: Add CI workflow for lint + typecheck on push/PR
I noticed that the repo has well-structured quality scripts (`biome check` via `bun run check`, and `check-types` via `tsc --noEmit`), but they only run locally — the only GitHub Actions workflow is `release.yml`.

A lightweight CI workflow that runs these checks on push/PR to `main` would catch lint and type errors before merge. Something like:

```yaml
name: CI
on:
  push: { branches: [main] }
  pull_request: { branches: [main] }
jobs:
  quality:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: oven-sh/setup-bun@v2
      - run: bun install --frozen-lockfile
      - run: bun run check
      - run: bun run check-types
```

Happy to submit a PR if this would be useful.

nnhhoang:
## Summary
- Adds a CI workflow that runs `bun run check` (Biome lint + format) and `bun run check-types` (TypeScript) on push/PR to `main`
- Uses concurrency groups to cancel stale runs
- Matches existing repo conventions from `release.yml`

Closes #148

## Test plan
- [ ] Push a PR to verify the workflow triggers
- [ ] Confirm `bun run check` and `bun run check-types` pass in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,28 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  quality:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: oven-sh/setup-bun@v2
+
+      - name: Install dependencies
+        run: bun install --frozen-lockfile
+
+      - name: Lint & format check
+        run: bun run check
+
+      - name: Type check
+        run: bun run check-types


Aymericr:
Hey @nnhhoang — thanks for putting this together, and apologies for sitting on it.

Closing this one, but not because the idea isn't right — the linked issue (#148) was closed as a duplicate of #147, which is still open. The real reason is that, in the time this sat, both `bun run check` and `bun run check-types` have drifted and currently fail on `main` (7 lint errors + ~32 type errors). Merging this workflow as-is would immediately block every push and PR until those are sorted, so the cleaner path is to land the cleanup first and then add CI.

We'll plan to address #147 ourselves once main is green again. Really appreciate the contribution though — the workflow you wrote is exactly the shape we'd reach for.

PR #180, opened 2026-03-26
fix: allow deleting walls and slabs via floating action menu

### LINKED ISSUE #156: No way to delete structures? (walls, slabs?)
Very cool project, just starting to play around with it...

One thing I noticed is I don't seem to be able to delete structures, maybe I'm missing it.

There's no overlay to move or delete for walls, slabs, etc...   
Doors do have the overlay, but attempting to move a door crashed the app (go back or reload the editor) 

There's a delete option in the toolbar, I toggled that on, but it also doesn't work.

Feel free to close if this is WIP, I know it's still early days.

nnhhoang:
## Summary
- Walls and slabs were excluded from `ALLOWED_TYPES` in `FloatingActionMenu`, so no delete UI appeared when selecting them
- Added `wall` and `slab` to the allowed types with a new `DELETE_ONLY_TYPES` list
- Made `onMove`/`onDuplicate` optional in `NodeActionMenu` so structural elements show only the delete button
- Move/duplicate remain available for items, doors, windows, roofs as before

Closes #156

## Test plan
- [x] Draw a wall, select it — floating menu shows delete-only button
- [x] Existing behavior preserved (doors/items still show move/duplicate/delete)
- [x] Lint passes (`bun run check` — 0 issues)
- [x] No new type errors introduced

### DIFF
--- a/packages/editor/src/components/editor/floating-action-menu.tsx
+++ b/packages/editor/src/components/editor/floating-action-menu.tsx
@@ -26,7 +26,8 @@ const DELETE_ONLY_TYPES = ['wall', 'slab']
 export function FloatingActionMenu() {
   const selectedIds = useViewer((s) => s.selection.selectedIds)
   const nodes = useScene((s) => s.nodes)
-  const deleteNode = useScene((s) => s.deleteNode)
+  const mode = useEditor((s) => s.mode)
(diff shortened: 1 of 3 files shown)


kleenkanteen:
@nnhhoang you should add a vid of the full flow btw

PR #190, opened 2026-03-27
 Improvements for GLTF loading and scene registry

korvixhq:
## What does this PR do?

Performance improvements for GLTF loading and scene registry:

- Added `useCachedGLTF` hook that caches loaded GLTFs by URL, eliminating redundant network requests when multiple items use the same asset.
- Removed redundant scene.traverse() call per item mount (previously ran once per item even with identical assets)
- Fixed registry cleanup null check for edge cases

## How to test

1. Open network tab and load a scene with multiple items of the same type (e.g., 10+ identical chairs)
2. Verify only 1 network request is made per unique GLTF asset
3. Place, move, and delete items to ensure no crashes or visual glitches occur with the cached GLTF and registry cleanup

## Screenshots / recordings

n/a

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -36,6 +36,7 @@ Node storage is a flat dictionary (`nodes: Record<id, AnyNode>`). Systems are pu
 
 - **components/viewer/** — Root `<Viewer>` canvas, camera, lights, post-processing, selection manager
 - **components/renderers/** — One renderer per node type (`WallRenderer`, `SlabRenderer`, …), dispatched by `NodeRenderer` → `SceneRenderer`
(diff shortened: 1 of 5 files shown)


Aymericr:
Thanks for the early performance work. Current main has replaced this surface: item loading now uses ItemGLTFLoader with explicit cache/failure cleanup (#480/#505), authored material capture and slot resolution must be preserved, and sceneRegistry.byType is a lazy Proxy so the cleanup guard is no longer needed.

Applying this old hook would mutate shared cached scenes/materials and bypass the current item material/export contracts. I’m closing it as superseded; any remaining GLTF performance issue should be reported against the current loader with a focused profile or reproduction.

PR #231, opened 2026-04-14
feat: Add catalog-based material presets across structural nodes

sudhir9297:
Added a shared, catalog-based material ( 13 items ) preset system across `wall`, `slab`, `roof`, `roof-segment`, `stair`, `stair-segment`, `ceiling`, and `fence`.

Key updates:
- node-type filtered material picker
- `custom` material fallback
- preset refs stored as `library:<id>`
- viewer resolves presets via `maps + mapProperties`
- textured materials now supported in the shared flow

Rendering/runtime fixes:
- async + cached texture loading to prevent rapid-switch crashes
- correct color-space handling for albedo vs data maps
- AO support for wall and fence UVs
- walls/slabs keep their real material instead of blue highlight overrides
- fixed post-processing fallback so it doesn’t render before WebGPU init


open-pascal (commented):
Rechecked the follow-up commits against the two regressions I called out earlier.

- The door/window material panel is no longer blank: `MaterialPicker` now keeps a custom fallback even when there are no catalog items.
- Visible wall selection/delete highlighting is restored in `wall-cutout.tsx`, and the highlighted variants now preserve the underlying wall material/preset appearance.

So the two blockers from my earlier review look addressed from my side. One smaller note remains: door/window are effectively custom-only now, instead of exposing the old preset swatches, but that is separate from the blank-state regression I flagged.

PR #235, opened 2026-04-15
fix: scene robustness (remount, empty geometry, asset URLs, first-person camera)

b9llach:
## What does this PR do?

A bundle of four small, defensive fixes to the viewer and editor that I hit while stress-testing a scanned-scene workflow on top of Pascal. Each is self-contained — happy to split this into four separate PRs if reviewers prefer, just let me know.

### 1. Renderer `markDirty` on `<Viewer>` remount (door / window / slab / ceiling)

When `<Viewer>` unmounts and remounts (preview mode toggle, view-mode switch, split-view toggle, etc.), these four renderers get stuck with their `<boxGeometry args={[0, 0, 0]}>` placeholder forever:

1. Renderer mounts, registers via `useRegistry`, renders the zero-size placeholder.
2. The corresponding system (DoorSystem, WindowSystem, SlabSystem, CeilingSystem) only processes nodes in `useScene.dirtyNodes`.
3. On remount the dirty set is empty — no one re-added the node after the previous unmount — so the system never regenerates the real geometry.
4. The renderer keeps the 0×0×0 placeholder until the user edits the node manually.

`WallRenderer` already handles this with a `useLayoutEffect` that marks its own node dirty on mount. This PR applies the same pattern to the four other renderers.

### 2. Defensive empty-geometry guard in `WallSystem`

When `generateExtrudedWall` returns an empty `BufferGeometry` (wall `start === end`, or a pathological mitering input), the previous `updateWallGeometry` assigned it to `mesh.geometry`, and the WebGPU renderer crashed in its next draw reading `.count` on `geometry.attributes.position` (undefined). Now we dispose the empty geometry, hide the mesh, and return early. The wall stays in the scene graph so Ctrl+Z can recover it.

### 3. `ItemRenderer`: guard `useGLTF` from empty URL + opaque placeholder

Two related issues:

- `useGLTF(resolveCdnUrl(node.asset.src) || '')` fell back to the empty string when `resolveCdnUrl` returned null (e.g. for `asset://` URLs, which need the async `resolveAssetUrl`). `useGLTF('')` resolves to the current page URL, gets HTML back, and GLTFLoader's JSON parser crashes with an unrecoverable error that Suspense refuses to retry (re-throwing the cached promise each render). Now we short-circuit on an unresolvable URL and render a solid placeholder instead.
- The `PreviewModel` Suspense fallback uses a material with `depthTest: false` and an animated time-based `opacityNode`. When it was used as the *permanent* render for items with no GLTF at all, the furniture pulsed and rendered through walls. Added a dedicated `PlaceholderBox` with a solid opaque `MeshStandardNodeMaterial` so item nodes without a model look like plain grey boxes that respect depth correctly.

### 4. `CustomCameraControls` returns null during editor first-person mode

`isFirstPersonMode` in `useEditor` mounts `<FirstPersonControls />` (pointer lock + WASD) as a sibling inside the viewer, but `CustomCameraControls` was still rendering drei's `<CameraControls>` alongside it because it only bailed on `walkthroughMode` (the viewer-package state, which the editor button doesn't flip). Two controllers fighting for the same camera meant the "walkthrough" button on desktop appeared to do nothing. Now `CustomCameraControls` returns `null` when `isFirstPersonMode === true`, leaving `FirstPersonControls` as the only camera driver.

## How to test

1. `bun dev`, open any scene with walls, doors, windows, slabs, and ceilings.
2. **Renderer remount fix:** toggle preview mode on and off, or swap between 3D / 2D / split view modes. Every element should remain visible. On `main`, doors / windows / slabs / ceilings vanish until you touch them.
3. **Wall empty-geometry guard:** defensive; not easily reachable via the UI (would need a wall with `start === end`). Exercised by manually calling `updateNode` on a wall to make its `end` equal to its `start` — the mesh hides instead of crashing the scene.
4. **Item empty-URL guard:** defensive; exercised by an `ItemNode` whose `asset.src` is `undefined` or `asset://…`. Instead of crashing on the HTML-is-not-JSON error, the item renders as a plain grey box that respects depth and sits correctly behind walls.
5. **First-person camera fix:** click the walkthrough / first-person button in the viewer toolbar on desktop; the camera should drop to eye height and WASD + mouse look should work. On `main`, the camera stays in orbit mode because drei's `CameraControls` is still winning the race with `FirstPersonControls`.

N/A — all four are regression / defensive fixes, most visible as "things that used to disappear or crash on `main` now don't."

## Checklist

- [x] My code follows the existing code style (`bun check` passes on the touched files — verified via `biome check` at `@biomejs/biome@^2.4.6` matching the repo's version)

### DIFF
--- a/packages/core/src/systems/wall/wall-system.tsx
+++ b/packages/core/src/systems/wall/wall-system.tsx
@@ -121,14 +121,32 @@ function updateWallGeometry(wallId: string, miterData: WallMiterData) {
 
   const newGeo = generateExtrudedWall(node, childrenNodes, miterData, slabElevation)
 
+  // Defensive: if the wall collapsed to zero length (bad data, or a
+  // cluster pass that over-merged short walls), `generateExtrudedWall`
+  // returns an empty BufferGeometry with no position attribute. The
+  // WebGPU renderer crashes reading `.count` on undefined, so hide the
+  // mesh instead of assigning the empty geometry. The wall stays in
+  // the scene graph (so Ctrl+Z can still recover it) but draws
+  // nothing until the start/end become valid again.
+  if (!newGeo.attributes.position) {
+    newGeo.dispose()
+    mesh.visible = false
+    return
+  }
+  mesh.visible = node.visible ?? true
+
   mesh.geometry.dispose()
   mesh.geometry = newGeo
   // Update collision mesh
   const collisionMesh = mesh.getObjectByName('collision-mesh') as THREE.Mesh
   if (collisionMesh) {
     const collisionGeo = generateExtrudedWall(node, [], miterData, slabElevation)
-    collisionMesh.geometry.dispose()
-    collisionMesh.geometry = collisionGeo
+    if (collisionGeo.attributes.position) {
+      collisionMesh.geometry.dispose()
+      collisionMesh.geometry = collisionGeo
+    } else {
+      collisionGeo.dispose()
+    }
   }
 
   mesh.position.set(node.start[0], slabElevation, node.start[1])
--- a/packages/editor/src/components/editor/custom-camera-controls.tsx
+++ b/packages/editor/src/components/editor/custom-camera-controls.tsx
@@ -22,6 +22,7 @@ const DEBUG_MAX_POLAR_ANGLE = Math.PI - 0.05
 export const CustomCameraControls = () => {
   const controls = useRef<CameraControlsImpl>(null!)
   const isPreviewMode = useEditor((s) => s.isPreviewMode)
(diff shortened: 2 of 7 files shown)


b9llach:
Closing — superseded by #241, which reapplies the same four defensive fixes on top of current main. #235 was opened against `f25806e` and became unmergeable after #231 (catalog-based material presets) and #236 (curved walls and fixes) touched adjacent code in `wall-system.tsx`, `slab-renderer.tsx`, and `ceiling-renderer.tsx`. Cleaner to open fresh than to rebase through two large merged commits.

PR #238, opened 2026-04-15
fix(editor): unit-aware sidebar controls + preserve user-typed precision

b9llach:
## What does this PR do?

Two related fixes to the sidebar's length inputs.

### 1. `SliderControl` is unit-aware

`useViewer.unit` (`'metric' | 'imperial'`) already existed and was toggled by the toolbar m/ft button, but the only places that actually respected it were `wall-measurement-label`, `site-edge-labels`, `metric-control`, and the 2D `floorplan-panel`. Every sidebar slider hard-coded `unit="m"` as its suffix and showed raw metres regardless of the user's preference. This PR teaches `SliderControl` to read `useViewer.unit` and, when the caller passes `unit="m"`, convert the value for display + editing while keeping scene data in metres (conversion is display-only). About 70 slider instances across every panel pick up imperial support with **zero call-site changes**.

Sliders with non-`"m"` unit strings (e.g. `"%"`, `"°"`) are unchanged.

### 2. Parent-side `Math.round(...)` clipping removed from panels

Every panel had a `value={Math.round(node.X * 100) / 100}` pattern around its sliders. It was there to defend against floating-point display noise (`2.10000000000000001` → `"2.10"`), but `SliderControl`'s own `.toFixed(precision)` already does that — so the parent rounds were redundant even in metric mode.

With the imperial path landed, the parent rounds became actively harmful. Tracing the door-height `precision={2}` case:

1. User sees `"8.01 ft"` (whatever was stored), types `8.00`, Enter
2. `submitValue` converts: `8 / 3.28084 ≈ 2.43840 m`
3. `onChange(2.43840)` — handler calls `updateNode({ height: 2.43840 })`
4. React re-renders; `door-panel.tsx` passes `value={Math.round(node.height * 100) / 100}` = `2.44`
5. `SliderControl` gets `2.44`, displays `toDisplay(2.44).toFixed(2)` = `(2.44 × 3.28084).toFixed(2)` = `"8.01"`

So typing `8.00` bounced back to `8.01`. Removing the `Math.round` in the parent lets the full-precision metres value survive the round trip, and the display cleanly shows `"8.00"` again.

The `submitValue` path also now rounds in the **display unit** before converting back to metres, so user-typed values preserve their exact precision regardless of which unit is stored underneath.

### New shared helper

`packages/editor/src/lib/units.ts` — `formatLength`, `formatArea`, `metersToFeet`, `feetToMeters`, `METERS_TO_FEET`. The slab + ceiling Info sections now use `formatArea(area, unit)` instead of a hardcoded `"m²"` suffix. Follow-up opportunity (out of scope for this PR): consolidate the four existing ad-hoc copies in `wall-measurement-label.tsx`, `site-edge-labels.tsx`, `metric-control.tsx`, and `floorplan-panel.tsx` onto this new module.

## How to test

1. `bun dev`
2. Click the m/ft toggle in the viewer toolbar.
3. Click a door: every slider suffix should flip from `m` to `ft`, and the numbers should reflect the conversion (2.1 m height shows as 6.89 ft).
4. Type `8.00` into any length slider while in imperial mode, press Enter: value should stay at `8.00 ft` (on `main` it snaps to `8.01 ft`).
5. Flip back to metric: everything returns to metres, no precision loss.
6. Click a slab, check the Info section: area should read `"12.50 m²"` in metric and `"134.55 ft²"` in imperial.

N/A — unit-toggle behaviour, hard to screenshot without video. Will record a short clip if reviewers want one.

## Checklist

- [x] My code follows the existing code style (`bun check` passes on the touched files — verified via `biome check` at `@biomejs/biome@^2.4.6`, which is the version pinned in root `package.json`)

### DIFF
--- a/packages/editor/src/components/ui/controls/slider-control.tsx
+++ b/packages/editor/src/components/ui/controls/slider-control.tsx
@@ -1,7 +1,9 @@
 'use client'
 
 import { useScene } from '@pascal-app/core'
+import { useViewer } from '@pascal-app/viewer'
 import { useCallback, useEffect, useRef, useState } from 'react'
+import { feetToMeters, metersToFeet } from '../../../lib/units'
 import { cn } from '../../../lib/utils'
 
 interface SliderControlProps {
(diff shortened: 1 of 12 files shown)


b9llach:
Closing — superseded by #248, which reapplies the unit-aware sidebar fix on current main. #238 became unmergeable after #231 (catalog-based material presets) rewrote adjacent panel code. The new PR is reduced to the two essential files (slider-control + units helper) since #231's panel restructuring already addressed the Math.round areas.

PR #252, opened 2026-04-16
feat(wall-tool): add Ctrl key to bypass grid snap for arbitrary wall lengths

### LINKED ISSUE #242: Remove the 0.5m snapping (even in 2D)
### Problem or motivation

Trying to trace a floor plan by uploading a .png to copy. However the snapping to 0.5m makes it impossible to trace accurately enough (even for rough sketches). 

It would be great to be able to turn off snapping or change the grid size.

### Proposed solution

_No response_

### Alternatives considered

_No response_

### Additional context

_No response_

tryflowk:
Closes #242

## Summary

- Adds `Ctrl` modifier key to the wall drawing tool that disables the 0.5m grid snap, allowing walls at any length (e.g. 1.70m, 2.35m)
- `Shift` continues to work as before — disabling angle snapping for non-45° walls
- Wall-join snapping (connecting to existing wall endpoints) still works in both free modes
- The two modifiers are fully independent and can be combined

## Modifier key behaviour

| Keys | Grid snap | Angle snap |
|------|-----------|------------|
| *(none)* | 0.5m grid | 45° |
| **Shift** | 0.5m grid | free |
| **Ctrl** | free | 45° |
| **Shift + Ctrl** | free | free |

## Test plan

- [ ] Draw a wall without any modifier — snaps to 0.5m increments and 45° angles as before
- [ ] Hold **Shift** while drawing — angle is free, grid snap still active
- [ ] Hold **Ctrl** while drawing — can produce walls like 1.70m, 2.35m, etc.
- [ ] Hold **Shift + Ctrl** — fully free movement
- [ ] Verify wall-join snap (connecting to existing wall endpoints) still works with Ctrl held

### DIFF
--- a/packages/editor/src/components/tools/wall/wall-drafting.ts
+++ b/packages/editor/src/components/tools/wall/wall-drafting.ts
@@ -99,10 +99,18 @@ export function snapWallDraftPoint(args: {
   walls: WallNode[]
   start?: WallPlanPoint
   angleSnap?: boolean
+  freeSnap?: boolean
   ignoreWallIds?: string[]
 }): WallPlanPoint {
-  const { point, walls, start, angleSnap = false, ignoreWallIds } = args
-  const basePoint = start && angleSnap ? snapPointTo45Degrees(start, point) : snapPointToGrid(point)
+  const { point, walls, start, angleSnap = false, freeSnap = false, ignoreWallIds } = args
+  let basePoint: WallPlanPoint
+  if (freeSnap) {
+    basePoint = point
+  } else if (start && angleSnap) {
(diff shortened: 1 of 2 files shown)


Aymericr:
Closing as obsolete against the current interaction contract. Snapping is now a persistent per-context mode: Shift cycles Grid/Lines/Angles/Off, Ctrl cycles the visible grid step (down to centimeter precision), and Alt owns force/free behavior. Reassigning Ctrl as a held wall-only bypass would conflict with that shared model; arbitrary placement is handled by the Off/Lines modes instead. Thank you for surfacing the original precision problem.

PR #271, opened 2026-04-22
Feat/color picker tool

sudhir9297:
## What does this PR do?

This PR introduces a new **Material Paint mode** that enables fast, direct material application across supported scene elements using a unified, click-based workflow.

It provides:
- A dedicated paint mode integrated into existing editor controls and shortcuts
- Surface-level material painting for walls, roofs, stairs, and other structural elements
- A reusable material system shared with existing panel-based editors
- A streamlined UI for selecting and applying materials without switching tools
- Performance-friendly interaction design for smooth painting workflows

---

### Material Paint Mode

- Added a new **Material Paint mode** in the main bottom controls alongside existing scene tools.
- Integrated paint mode with:
  - Keyboard shortcuts
  - Command palette entry points
- Reused the existing material library/picker (no separate paint-only system).
- Made paint mode context-aware:
  - Material list updates based on the selected mesh type (wall, roof, stair, etc.).
- Enabled persistent painting:
  - Users can apply materials continuously without exiting the tool.

---

### Material System Integration

- Added shared material-paint helpers so paint mode uses the same logic as panel-based material editing.
- Kept wall, roof, and stair panel material flows in sync with the active paint material.
- Ensured consistent behavior between painting and manual panel edits.

---

### Paint UI & Interaction

- Built a **paint tray** in the bottom UI displaying the material library for the active target.
- Updated material strip:
  - Horizontal scrolling layout
  - No visible scrollbars
  - Mouse-wheel horizontal scroll support
- Allowed independent resizing of the material strip from the bottom toolbar.
- Normalized material tile visuals for better consistency across light/dark textures.

---

### Interaction & Feedback

- Added dedicated paint cursor with floating badge for clear mode indication.
- Removed heavy hover-preview logic to avoid FPS drops during painting.
- Simplified invalid interaction feedback:
  - Uses a quick "not-allowed" cursor instead of custom blocked UI.
- Preserved selection context:
  - Entering paint mode uses the currently selected mesh to initialize target/material state.

---

### Performance & Stability

- Reduced runtime overhead by removing continuous hover-based preview updates.
- Ensured smooth interaction even when painting repeatedly across multiple surfaces.


wass08 (changes requested):
# PR #271 review — blocker fixes
 
## Core issue
 
Paint mode is a 100% editor feature — the read-only `/viewer/[id]` route has no paint tool and users in that route never change materials. The PR leaks paint-tool vocabulary into `packages/viewer` in two places (`materialPreview` state, `paint-ready`/`paint-disabled` hover modes) and also bypasses the package boundary with relative-path imports. All three blockers resolve together if we stop treating the viewer as a place where editor features can live.
 
---
 
## 1. Remove `materialPreview` from `useViewer`
 
**Files:** `packages/viewer/src/store/use-viewer.ts`, `use-viewer.d.ts`, and the eight `packages/viewer/src/components/renderers/*-renderer.tsx` files.
 
The `materialPreview` state, `setMaterialPreview`, `clearMaterialPreview`, and the `MaterialPreview` type never get written to anywhere in the codebase — grep confirms zero callers. The renderer branches that read it are permanently dead. Meanwhile the actual paint preview is implemented by directly swapping `mesh.material` via `sceneRegistry` in `selection-manager.tsx`, which is the right pattern per `.claude/rules/viewer-isolation.md` (editor injects behaviour into the canvas without touching the viewer package).
 
**What to do:** delete all of it. Revert every renderer to reading material from its node only. Keep the selection-manager's direct `mesh.material` swap — that approach stays.
 
---
 
## 2. Stop reaching into `../../../../viewer/src/…` from the editor
 
**File:** `packages/editor/src/components/editor/selection-manager.tsx:32-41`
 
The paint preview needs wall/roof/stair material helpers, but the current imports bypass `@pascal-app/viewer` entirely:
 
```ts
import { applyMaterialPresetToMaterials, createMaterial, createMaterialFromPresetRef } from '../../../../viewer/src/lib/materials'
import { getRoofMaterialArray } from '../../../../viewer/src/systems/roof/roof-materials'
import { getStairBodyMaterials, getStairRailingMaterial } from '../../../../viewer/src/systems/stair/stair-materials'
import { getVisibleWallMaterials } from '../../../../viewer/src/systems/wall/wall-materials'
```
 
**What to do:** export those six helpers from `packages/viewer/src/index.ts` and import them through `@pascal-app/viewer` like any external consumer. The editor then lives on the package's public API only.
 
---
 
## 3. Generalize `hoverHighlightMode` — opaque name + styles prop + uniform-driven rendering
 
**Files:** `packages/viewer/src/store/use-viewer.ts`, `packages/viewer/src/components/viewer/post-processing.tsx`, and the editor-side `selection-manager.tsx` effect that sets the mode.
 
Two problems compound here:
 
First, the viewer bakes editor mode names into its type (`'default' | 'delete' | 'paint-ready' | 'paint-disabled'`). The viewer has no business knowing those names exist.
 
Second, the post-processing pass currently memoizes the entire TSL node graph on `hoverHighlightMode`. Every mode change rebuilds the outline pass and retriggers a shader compile. Colors and strength should be uniforms that update reactively without touching the graph.
 
**What to do:**
 
Change `useViewer.hoverHighlightMode` to an opaque `string`. The viewer doesn't care what the name means.
 
Add a `hoverStyles` prop on `<Viewer>` mapping names to visual style. Ship a sensible default so the read-only route works untouched:
 
```tsx
type HoverStyle = {
  visibleColor: number   // 0xRRGGBB
  hiddenColor: number
  strength: number       // ~1–10
  pulse: boolean
}
 
<Viewer
  hoverStyles={{
    default:          { visibleColor: 0x00aaff, hiddenColor: 0xf3ff47, strength: 5, pulse: true },
    delete:           { visibleColor: 0xef4444, hiddenColor: 0x991b1b, strength: 6, pulse: false },
    'paint-ready':    { visibleColor: 0xf59e0b, hiddenColor: 0xfde068, strength: 5, pulse: true },
    'paint-disabled': { visibleColor: 0x94a3b8, hiddenColor: 0x475569, strength: 4, pulse: false },
  }}
/>
```
 
In `post-processing.tsx`, build uniforms once and drive them from an effect — no `useMemo` deps on `hoverHighlightMode`:
 
```ts
const visibleColor  = useMemo(() => uniform(new Color()), [])
const hiddenColor   = useMemo(() => uniform(new Color()), [])
const strength      = useMemo(() => uniform(5), [])
const pulseEnabled  = useMemo(() => uniform(1), [])  // 0 = constant, 1 = oscillating
 
useEffect(() => {
  const s = hoverStyles[hoverHighlightMode] ?? hoverStyles.default
  visibleColor.value.setHex(s.visibleColor)
  hiddenColor.value.setHex(s.hiddenColor)
  strength.value = s.strength
  pulseEnabled.value = s.pulse ? 1 : 0
}, [hoverHighlightMode, hoverStyles])
```
 
Fold the pulse ternary into the graph so it's uniform-controlled too:
 
```ts
const oscillating = oscSine(time.div(pulsePeriod).mul(2)).mul(0.5).add(0.5)
const osc = mix(oscillating, float(1), pulseEnabled)  // pulse: 0 → animated, 1 → held
```
 
Now mode switches never rebuild the graph — four `.value` writes, done. The editor owns all mode names (`delete`, `paint-ready`, `paint-disabled`, future modes) and passes the corresponding table; zero editor vocabulary in the viewer package.
 
---
 
## Net result
 
`useViewer` shrinks back to the presentation vocabulary the rule describes: selection, camera/level/wall mode, theme, display toggles, plus `hoveredId` and an opaque `hoverHighlightMode` string. The `/viewer/[id]` route continues to work with zero changes. The editor keeps its current direct-material-swap preview mechanism, just wired through the public viewer API.

wass08 (changes requested):
Almost there — one last thing. `DEFAULT_HOVER_STYLES` in the viewer still names editor-specific modes (`delete`, `paint-ready`, `paint-disabled`), so the viewer package keeps low-grade knowledge of editor vocabulary that it's now explicitly trying not to have. The editor already passes its own `EDITOR_HOVER_STYLES` with the full table, so the viewer only needs the `default` to stay a safe fallback for read-only consumers.

**Change** — `packages/viewer/src/components/viewer/post-processing.tsx`:

```diff
 export const DEFAULT_HOVER_STYLES: HoverStyles = {
   default: DEFAULT_HOVER_STYLE,
-  delete: {
-    visibleColor: 0xef_44_44,
-    hiddenColor: 0x99_1b_1b,
-    strength: 6,
-    pulse: false,
-  },
-  'paint-ready': {
-    visibleColor: 0xf5_9e_0b,
-    hiddenColor: 0xfd_e0_68,
-    strength: 5,
-    pulse: true,
-  },
-  'paint-disabled': {
-    visibleColor: 0x94_a3_b8,
-    hiddenColor: 0x47_55_69,
-    strength: 4,
-    pulse: false,
-  },
 }
```

Nothing else moves — `EDITOR_HOVER_STYLES` in `packages/editor/src/components/editor/index.tsx` already carries the full table and is already passed to both `<Viewer>` mounts. After this, grepping the `viewer` package for `delete`, `paint-ready`, or `paint-disabled` should return zero hits; the viewer becomes fully agnostic to any mode beyond `default`, and the editor owns the complete name → style map on its side.

---

That's the whole fix. Blocks merge on this repo since the intent of the whole refactor is to keep editor vocabulary out of `packages/viewer` — shipping editor mode names in the viewer's default data contradicts that.

wass08 (approved):

PR #273, opened 2026-04-22
Add a minimal Bun devcontainer setup

elalemanyo:
## Summary
- Add a minimal `.devcontainer` config so the repo can be opened and run in Bun without extra setup
- Forward the editor port used by the app and install dependencies on container creation
- Update the local dev docs to match the actual dev server port

## Why
This makes onboarding easier and gives contributors a repeatable local environment with less machine-specific setup.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,11 @@
+{
+  "name": "Pascal Editor",
+  "image": "oven/bun:1.3.0",
+  "features": {
+    "ghcr.io/devcontainers/features/git:1": {},
+    "ghcr.io/devcontainers/features/github-cli:1": {}
+  },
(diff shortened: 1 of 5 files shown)


Aymericr:
Keeping this open as the developer-container track (separate from the production Docker PR). Please rebase onto current main, use bun install --frozen-lockfile, and verify bun dev exposes the editor on 3002 from a clean container. Limit the refreshed diff to .devcontainer plus any still-incorrect current docs; the old port corrections may already have landed independently.

Aymericr:
Thank you for the contribution. I ported this onto current main in #520, preserved your original commit authorship, corrected the stale 3000 references, added frozen-install and Dev Container feature locking, and verified the container end to end through an HTTP 200 on port 3002. Closing this PR as superseded by the production-ready continuation.

PR #325, opened 2026-05-21
fix: update mcp and nodes for SiteNode.children string[] schema change

jelharou:
## What does this PR do?

Fixes breakage introduced by #320, which changed `SiteNode.children` from embedded `BuildingNode | ItemNode` objects to flat `string[]` IDs. Three call sites in `packages/mcp` and `packages/nodes` were not updated.

- **`mcp/lib/rehydrate-site-children.ts`** — replace now-obsolete re-embedding logic with a passthrough (call-site compatible, callers become no-ops)
- **`mcp/tools/variants/generate-variants.ts`** — drop the inline copy of the same function and its call
- **`nodes/src/site/renderer.tsx`** — cast `childId as AnyNodeId` since `SiteNode.children` is now `string[]`

## How to test

1. `bun dev` — confirm the build completes with no TS errors in `@pascal-app/mcp` and `@pascal-app/nodes`
2. Open the editor, load a scene with a site node — confirm site children render correctly
3. Run the `generate_variants` MCP tool — confirm variants produce valid node counts

N/A — non-visual fix (type correctness and schema alignment).

## Checklist

- [x] My code follows the existing code style (run \`bun check\` to verify)

### DIFF
--- a/packages/mcp/src/lib/rehydrate-site-children.ts
+++ b/packages/mcp/src/lib/rehydrate-site-children.ts
@@ -1,40 +1,12 @@
 import type { SceneGraph } from '@pascal-app/core/clone-scene-graph'
-import type { AnyNode } from '@pascal-app/core/schema'
 
 /**
- * `cloneSceneGraph` normalises `SiteNode.children` to an array of node IDs,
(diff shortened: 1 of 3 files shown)


wass08:
Thank you!

PR #329, opened 2026-05-21
fix: resolve 8 test failures (stair fixtures, spawn icon, Zod v4 introspection, three-mesh-bvh)

jelharou:
## What does this PR do?

Fixes all 8 test failures that appeared after pulling the latest `main` (shelf v2, IconRef url kind, StairOpeningSystem commits). Four distinct root causes:

**1. `stair-opening-sync` tests (4 failures)**
`manualOpening` fixtures used `z ∈ [0.8, 2.9]` but the computed stair opening expands to `z ∈ [0.05, 2.95]` (`position[2]=0.2` minus `openingOffset=0.15`). `isCoveredByExistingHole` requires full polygon containment, so the auto-hole was not suppressed. Fixtures updated to `z ∈ [0, 3]`.

**2. Spawn parity test (1 failure)**
Commit `6a1d853` (IconRef: add `url` kind) changed all registered icon refs from `{ kind: 'iconify' }` to `{ kind: 'url' }`. Test assertion still expected `'iconify'`. Updated to `'url'`.

**3. `builtinPlugin` discriminator test (1 failure)**
`nodeType()` wraps `z.literal(kind)` in `ZodDefault` so nodes parse without an explicit `type` field. In Zod v4, `ZodDefault` has no `.value` property — the test was written for Zod v3's bare `ZodLiteral`. Fixed introspection to unwrap via `.def?.innerType` before reading `.value`.

**4. `geometry.test.ts` errors (2 errors)**
Shelf v2 added `import … from '@pascal-app/viewer'` to `geometry.ts`. The viewer barrel transitively loads `wall-system` → `three-bvh-csg` → `three-mesh-bvh@0.9.9` as CJS/UMD (because `three-bvh-csg` lacks `exports` conditions). That UMD has a class-initialization-order bug in Bun's test environment. Added `bunfig.toml` + `test-preload.ts` to mock `three-bvh-csg` globally — no test exercises CSG at runtime, so this is safe.

## How to test

1. Pull this branch: `git fetch fork fix/test-failures-after-pull && git checkout fix/test-failures-after-pull`
2. Run `bun test` from the repo root
3. Expect: **662 pass, 0 fail, 0 errors**

N/A — non-visual change (test fixes only).

## Checklist

### DIFF
--- a/bunfig.toml
+++ b/bunfig.toml
@@ -1,0 +1,2 @@
+[test]
+preload = ["./test-preload.ts"]
--- a/packages/core/src/systems/stair/stair-opening-sync.test.ts
+++ b/packages/core/src/systems/stair/stair-opening-sync.test.ts
@@ -78,11 +78,14 @@ describe('syncAutoStairOpenings', () => {
     const building = BuildingNode.parse({ name: 'Building' })
     const ground = LevelNode.parse({ name: 'Ground', level: 0, parentId: building.id })
     const upper = LevelNode.parse({ name: 'Upper', level: 1, parentId: building.id })
+    // Opening must fully contain the computed stair polygon including the 0.15 m
+    // openingOffset expansion. Stair at position [2,0,0.2], width=1, length=2.6
+    // → expanded opening z ∈ [0.05, 2.95]. Use z ∈ [0, 3] with margin.
     const manualOpening: Array<[number, number]> = [
-      [1.2, 0.8],
-      [2.8, 0.8],
-      [2.8, 2.9],
-      [1.2, 2.9],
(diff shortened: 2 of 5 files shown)


Aymericr:
Thanks for tackling the test failures! `main` has moved a fair bit since this was opened and the failing set is now different (the current reds are around SSRF IPv6 handling, `createWallOnCurrentLevel`, and elevator-selection candidates), so the fixtures touched here no longer line up and the PR conflicts. Closing for now — but a fresh PR targeting the current failures would be very welcome, and we're doing a maintainer pass on those too. Appreciate the effort!

PR #330, opened 2026-05-22
feat(roof-system): six roof-accessory kinds (chimney, dormer, skylight, solar-panel, ridge-vent, box-vent) on the registry model

sudhir9297:
## What does this PR do?

Ports the six roof-mounted accessories — **chimney, dormer, skylight, solar-panel, ridge-vent, box-vent** — to the registry-driven node model under `packages/nodes/src/<kind>/`, and extends the registry contract so the framework code never has to name them by kind.

Beyond the kind ports, this branch contributes:

- **Three new `NodeDefinition` slots** that took the per-kind branches out of framework code:
  - `capabilities.roofAccessory` — hosts the dirty-cascade to the parent roof + an optional `buildCut` so dormer / skylight contribute their CSG-cut geometry through the registry instead of being named inside `RoofSystem`.
  - `capabilities.paint` — `resolveRole` / `buildPatch` / `applyPreview` / `getEffectiveMaterial`. Chimney, dormer, and wall now route paint hover / click / preview / picker through this; per-kind arms deleted from `selection-manager.tsx` + `material-paint.ts`.
  - `keyboardActions` — `R` / `T` handlers contributed by the kind. Skylight's open/close moves from `editor/src/lib/skylight-interaction.ts` to `packages/nodes/src/skylight/interaction.ts`.
- **Roof-segment per-surface materials** (top / edge / wall paint, with parent-roof fallback) and the registry-driven floor-plan + Add-Element flow in the roof inspector.
- **Schema migrations**: roof-segment gains `pitch` + multi-slope shape ratios (and `getActiveRoofHeight` / `getPitchFromActiveRoofHeight` helpers in core); existing scenes are migrated in `useScene`'s loader.
- **Viewer cleanup**: the merged-roof CSG loop is now registry-driven (no more `if (childElem.type === 'skylight' || ...)`); the dormer-specific helpers (`generateDormerGeometry`, `buildDormerCutBrush`, etc.) moved out of `packages/viewer/src/systems/roof/roof-system.tsx` into the dormer kind, leaving only roof-segment-generic primitives (`getRoofSegmentBrushes`, `getRoofOuterSurfaceFrameAtPoint`, `roofCsgDummyMats`, `mapRoofGroupMaterialIndex`, `remapRoofShellFaces`) in viewer.

### Follow-ups left intentionally

- Roof / stair paint still use editor-side arms (segment→parent fan-out needs `buildPatch` to return `{ targetId, patch }` — separate contract change).
- Door / window interaction (`R` / `T` open/close) still lives in `editor/src/lib/` because `first-person-controls.tsx` invokes the helpers with `{ persist: false }`, which `KeyboardAction.run(node)` doesn't expose yet.

## How to test

1. **Place each accessory.** From the roof inspector's "Add Element" section, drop one chimney, one dormer, one skylight, one solar-panel, one ridge-vent, and one box-vent onto a selected roof. Move them in 3D and confirm the host roof re-CSGs cleanly (no stale cut after the drag commits).
2. **Roof-mounted CSG.** Resize a dormer or skylight with the inspector sliders; confirm the merged roof shell rebuilds the cut on commit and the dormer body sits flush against the slope.
3. **Per-segment paint.** Enter segment-edit mode on the roof, open the Material picker, and paint a single segment's top / edge / wall — only that segment changes. Exit segment-edit and paint the merged roof — the unpainted segments inherit, painted segments keep their override.
4. **Chimney / dormer paint.** Click body vs. cap on a chimney with paint mode active; both write to the right field. Same for dormer wall / side / top.
5. **Wall paint via registry.** Paint an interior wall surface, then an exterior one. Both should preview correctly and commit. (This change routes wall through the new `capabilities.paint` dispatch — regression-test against `main`'s behaviour.)
6. **Skylight R / T.** Select an opening / sliding skylight and press R (toggle) then T (close); animation runs through `useInteractive.skylightAnimations`.
7. **Existing-scene migration.** Open a scene saved on `main` containing a legacy single-segment roof — the loader rewrites it into the segment-with-pitch shape, no visual regression.

```bash
bun typecheck             # all five packages clean
cd packages/nodes && bun test src/{box-vent,chimney,dormer,ridge-vent,skylight,solar-panel,wall}
# 84 pass / 0 fail
```

N/A — to be added before merge. The visual surface is the six accessory kinds and the roof inspector's Add-Element section; behaviour should be a strict superset of `main`.

## Checklist


Aymericr (commented):
## Review — six roof accessories on the registry model

**CI: ✅ | Tests: 84 pass / 0 fail**

### What's solid

The architecture is on point. All six kinds land under `packages/nodes/src/<kind>/` following the three-checkbox model, the registry wiring is correct (schema → bus → AnyNode union → builtinPlugin), and the viewer's merged-roof CSG loop is genuinely kind-agnostic now — no more `if (childElem.type === 'skylight' || ...)` checks. The three new `NodeDefinition` slots (`capabilities.roofAccessory`, `capabilities.paint`, `keyboardActions`) are clean abstractions that let future kinds contribute behaviour without touching framework code.

A few things to address before this is merge-ready:

---

### 🚫 Remove `.claude/PORT-CHEATSHEET.md`

```
.claude/PORT-CHEATSHEET.md — 263 lines
```

This is an AI-authoring reference for porting kinds — not useful to community contributors and doesn't belong in the public repo. Please drop it from the commit. If the canonical how-to is worth keeping, the right home is `wiki/architecture/` or a page in the contributing docs.

---

### 📄 Document the three new `NodeDefinition` capabilities

`registry/types.ts` adds `roofAccessory`, `paint`, and `keyboardActions` to `NodeDefinition`. The PR description explains each well, but `wiki/architecture/node-definitions.md` (referenced by the cheatsheet as the authority on the three-checkbox model) should be updated to reflect them — otherwise the next contributor porting a kind won't find them.

---

### 🖼️ Screenshots before merge

The checklist has "N/A — to be added before merge." For a PR that lands six visual 3D components, even a short screen recording of place → move → rotate → paint on each kind would help reviewers validate behaviour without needing to run the branch locally.

---

### Minor: double registry lookup in `use-keyboard.ts`

```ts
} else if (node && nodeRegistry.get(node.type)?.keyboardActions?.r?.appliesTo(node)) {
  e.preventDefault()
  nodeRegistry.get(node.type)!.keyboardActions!.r!.run(node)
```

The `nodeRegistry.get()` call is duplicated. Worth caching the ref:

```ts
const action = node ? nodeRegistry.get(node.type)?.keyboardActions?.r : undefined
if (action?.appliesTo(node!)) {
  e.preventDefault()
  action.run(node!)
  ...
}
```

Not a blocker, but the non-null assertions are noisy when the lookup can just be stored once.

---

### FYI: dormer geometry is a stub

I noticed the dormer definition comment says *"variant-specific dormer roof shapes, window opening + frame, sill, and the CSG trim where the dormer meets the host roof are deferred."* That's fine to ship — just worth tracking as a follow-up issue so it doesn't get lost. Is there an existing issue for this?

---

**Summary:** The structural work here is good and the registry pattern is applied correctly. Main asks: drop the Claude cheatsheet file, add docs for the three new capabilities, and get screenshots in before merge.

PR #351, opened 2026-05-30
chore(dev): gate react-scan overlay behind env flag

1pitaph:
## What does this PR do?

Removes the automatic `react-scan` CDN injection in development and makes the React render debugging overlay opt-in.

`react-scan` now only runs when `NEXT_PUBLIC_REACT_SCAN=true` is set in development. The opt-in flag is documented in `.env.example`, so local developers can enable it when debugging render performance without showing the overlay by default.

Also fixes sidebar collapse rendering instability in the editor:

- Removes the sidebar collapse width transition.
- Reads the latest sidebar store state when toggling, so rapid repeated clicks do not use stale collapsed state.
- Forces the viewer canvas size to sync immediately after layout changes.
- Ensures post-processing fallback renders use the active scene theme background instead of flashing black or white.

## How to test

1. Run `bun dev` without setting `NEXT_PUBLIC_REACT_SCAN`, open the editor, and verify the purple `react-scan` component overlay does not appear.
2. Add `NEXT_PUBLIC_REACT_SCAN=true` to `.env.local`, restart `bun dev`, refresh the page, and verify the overlay appears.
3. Remove the flag or set it to `false`, restart `bun dev`, and verify the overlay is disabled again.
4. Open the editor and repeatedly click the sidebar collapse / expand button.
5. Verify the viewer does not flash a white or black blank area during the resize.
6. Verify the main 3D view does not enter the black / white corrupted rendering state after repeated toggles.

N/A. This change removes a development-only debug overlay by default and fixes a transient interaction/rendering issue.

## Checklist

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -3,5 +3,8 @@
 # Google Maps API key (enables address search)
 # NEXT_PUBLIC_GOOGLE_MAPS_API_KEY=your_google_maps_api_key
 
+# React render debugging overlay (disabled by default)
+# NEXT_PUBLIC_REACT_SCAN=false
+
 # Dev server port (default: 3000)
 # PORT=3000
--- a/apps/editor/app/client-bootstrap.tsx
+++ b/apps/editor/app/client-bootstrap.tsx
@@ -11,9 +11,12 @@
 import '../lib/bootstrap'
 import { type ReactNode, useEffect } from 'react'
 
+const shouldEnableReactScan =
+  process.env.NODE_ENV === 'development' && process.env.NEXT_PUBLIC_REACT_SCAN === 'true'
+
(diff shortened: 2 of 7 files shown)


Aymericr:
The opt-in React Scan idea is sound, but this PR's head is a long-lived main branch and now carries unrelated layout, resize, and viewer changes against an obsolete base. It cannot be merged safely as the requested env gate.

I’m closing this stale snapshot. A fresh current-main PR limited to client-bootstrap.tsx and .env.example, with the overlay disabled unless both development mode and NEXT_PUBLIC_REACT_SCAN=true, would be straightforward to review.

PR #356, opened 2026-06-02
docs(mcp): document plan↔world coordinate convention

### LINKED ISSUE #337: Docs: document the plan↔world coordinate convention for MCP/API authors
## Summary

The MCP package documents units (metres) and per-tool parameters well, but I couldn't find an author-facing statement of the **coordinate convention**. When generating geometry programmatically (e.g. from a land survey), this is the first thing you need and currently have to reverse-engineer from source.

## What's missing

There's no doc stating that:
- 2-D points — `wall.start`/`wall.end` and the `polygon`/`holes` of `slab`, `zone`, `ceiling` — are `[x, z]` on the **X–Z ground plane**, with **Y up**.
- The second component maps **straight to world Z, no negation**: the editor builds geometry as `new Vector3(point[0], y, point[1])`.
- Rotations are **radians**, stored as Euler `[x, y, z]` tuples.

This matters for anyone authoring from a north-up source (a survey with bearings from North, a north-up site plan, most 2-D plotting libraries): because the plan embeds into X–Z, such layouts come in **mirrored** when viewed top-down, and there's currently no note about it. `examples/generate-apartment.md` only uses axis-aligned boxes (`[0,0] → [10,0]`), so the handedness never surfaces in any example.

## Suggestion

Add a short **"Coordinate conventions"** section to `packages/mcp/README.md`: axes, units, radians, the `[x,z] → (x, y, z)` mapping, the top-down-mirror caveat for externally-computed coordinates, and one non-axis-aligned example. Happy to open a PR if that's welcome.

## Verified against `pascalorg/editor` @ main

- `packages/core/src/services/snap.ts` — *"Snaps a 3D point to a regular grid in the X/Z plane, preserving Y."*
- `packages/core/src/services/movement.ts` — `movePlanToward` packs `[x, y, z]` and returns `[result[0], result[2]]`.
- `packages/nodes/src/{wall,slab,ceiling}/tool.tsx` — `new Vector3(point[0], y, point[1])`.
- `packages/core/src/schema/nodes/level.ts` — `level: z.number()` (floor base elevation, metres).

*(Found while building a site model from a real property report via the MCP server — the plan↔world handedness was the one thing that took real digging to pin down.)*

marcelgruber:
Adds a **Coordinate conventions** section to `packages/mcp/README.md`,
placed right before `## Tools` so authors read it before touching
geometry. Backed by a 71-node MCP demonstration scene that stress-tests
every claim the section makes.

Closes #337.

## Why

The README documents units (metres) and per-tool parameters well, but
there's no author-facing statement of the coordinate convention. When
generating geometry programmatically (e.g. from a land survey), this is
the first thing you need and currently has to be reverse-engineered from
source. The `examples/generate-apartment.md` walkthrough only uses
axis-aligned boxes (`[0,0] → [10,0]`), so the handedness never surfaces
in any existing example.

## What the new section covers

- Right-handed scene; X and Z form the ground plane, Y is up.
- Metres for lengths; radians for rotations, stored as Euler `[x, y, z]`.
- The `[x, z] → (x, y, z)` mapping for 2-D points (`wall.start` / `end`,
  `slab` / `zone` / `ceiling` `polygon` and `holes`) — no sign flip.
- **Heads-up for externally-computed coordinates:** Pascal's viewports
  apply their own rotations on top of the world axes. The 2-D plan
  panel wraps content in a +90° rotation (`FLOORPLAN_VIEW_ROTATION_DEG`
  in `floorplan-panel.tsx`), and the 3-D "top-down" snap defaults to a
  ~45° offset between world and screen axes when invoked from the iso
  default position. So a layout authored as if "Y = north, viewed
  top-down" will arrive **rotated** relative to its source — and
  possibly further reflected, depending on viewport state. The
  reliable fix is a scaled guide image at known anchors.
- One non-axis-aligned worked example (a 6 × 4 m slab rotated 30°)
  whose vertices verify analytically as sides 6/4/6/4 m, `AB · AD = 0`,
  heading 30.0007° from +X.

## How this got verified

I stress-tested the section against actual Pascal-generated geometry
before publishing. An iterative MCP-driven demo (see
[`examples/coordinate-conventions-demo.md`](../tree/docs/mcp-coordinate-conventions/packages/mcp/examples/coordinate-conventions-demo.md))
was authored, persisted, and inspected in the editor. The first
version of this PR claimed external coordinates "arrive mirrored when
viewed top-down" — but inspecting the rendered demo showed they arrive
**rotated**, not mirrored, and the proposed "reflect across the axis"
corrective doesn't help. The section and the demo were rewritten before
pushing. The demo is now included with the PR so future readers can
verify the same way.

### Demo files

- [`packages/mcp/examples/coordinate-conventions-demo.json`](../tree/docs/mcp-coordinate-conventions/packages/mcp/examples/coordinate-conventions-demo.json)
  — 71-node SceneGraph: reference compass at the origin, axis-aligned
  baseline (Demo A), rotated-30° example verbatim (Demo B, verified
  analytically), paired page-intent / world-result L pair (Demo C), and
  z-reflected variant showing the mis-corrective (Demo D), plus a
  takeaway band on the south edge. Loadable with
  `pascal-mcp --scene examples/coordinate-conventions-demo.json`.
- [`packages/mcp/examples/coordinate-conventions-demo.md`](../tree/docs/mcp-coordinate-conventions/packages/mcp/examples/coordinate-conventions-demo.md)
  — walks through what the scene is and what each demo proves.

### What Demo B verified

Loading and inspecting the polygon `[[0,0],[5.196,3.0],[3.196,6.464],[-2.0,3.464]]`:

```
side_lengths_m:               [5.9999, 3.9999, 5.9999, 3.9999]
first_edge_heading_deg_from_+X: 30.0007
AB · AD (perpendicularity):    0.0
```

So the example in the README is, to 4 decimal places, a 6 × 4 m rectangle
with its first edge at 30° CCW from +X and adjacent edges perpendicular.

### What Demo C / D showed about the doc

The L authored on a "north-up page" with page-+x right and page-+y up,
pasted directly as `[x, z]`, lands in Pascal's 2-D plan panel rotated
90° clockwise — stem along screen-right rather than screen-up. The
z-reflected variant (Demo D) lands as a *mirror* of Demo C, not as a
page-correct L. That's why the original draft's "reflect across the
appropriate axis" advice was wrong for Pascal's actual viewport
behavior, and why the section now points at viewport rotations as the
mechanism.

## Verified against `pascalorg/editor` @ main

- `packages/core/src/services/snap.ts` — *"Snaps a 3D point to a regular
  grid in the X/Z plane, preserving Y."*
- `packages/core/src/services/movement.ts` — `movePlanToward` packs
  `[x, y, z]` and returns `[result[0], result[2]]`.
- `packages/nodes/src/{wall,slab,ceiling}/tool.tsx` — `Vector3`
  construction uses `(plan[0], y, plan[1])` (destructured form:
  `([x, z]) => new Vector3(x, y, z)`).
- `packages/core/src/schema/nodes/level.ts` — `level: z.number()` (floor
  base elevation, metres).
- `packages/editor/src/components/editor/floorplan-panel.tsx` —
  `FLOORPLAN_VIEW_ROTATION_DEG = 90`, applied as
  `<g transform="rotate(${floorplanSceneRotationDeg})">` around the
  floor-plan content group.

## Out-of-scope but worth surfacing

While building the demo scene I hit a few API quirks in the hosted
`editor.pascal.app` MCP server (v0.6.1) that are unrelated to this
docs PR but a maintainer may want to track separately:

- `save_scene` with `graph` + `includeCurrentScene:false` tail-throws
  `MCP error -32600: undefined is not an object (evaluating 'this.loadProject')`
  even though the SceneStore row is durably written (the demo JSON
  here was published this way; `get_project_status` confirms the
  expected `nodeCount` and `graphHash`).
- `apply_patch` live-syncs against the same broken `loadProject` path;
  earlier deletes can persist while creates in the same batch roll
  back, so a batch is not atomic in practice.
- `create_project`'s `isPrivate: false` flag is declared in the input
  schema and the storage types, but the editor route still
  307-redirects anonymous visitors. So "public sharing" via the API
  alone doesn't currently produce an unauth-viewable URL.
- The renderer fails on zone names past ~500 chars (the editor shows
  "the editor scene failed to render"); shorter labels render fine.
  Bisected against several other suspects (small zones, concave
  half-scale slabs, thin slabs) — the name-length was the trigger.

Happy to break any of these out into separate issues if useful.

### DIFF
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -247,16 +247,29 @@ There is no sign flip. The editor builds plan geometry as
 the owning `level`'s elevation (its `level` field, in metres) plus the
 element's own height; slabs additionally carry an absolute `elevation`.
 
-**Heads-up when you compute coordinates outside the editor.** Because the plan
-is the X–Z plane, the second coordinate is **depth**, not a screen-"up" axis.
-If you generate geometry from a source that assumes *"Y = north, viewed from
-the top"* — a land survey with bearings from North, a north-up site plan, or
-most 2-D plotting libraries — the result can arrive **mirrored** relative to
-that source when viewed top-down. Pascal's own 2-D and 3-D tools are
-internally consistent, so this only affects geometry authored programmatically.
-Decide your axis mapping explicitly and verify orientation against a known
-reference (e.g. a scaled guide image) before trusting it; reflect across the
-appropriate axis if it comes in flipped.
+**Heads-up when you compute coordinates outside the editor.** Pascal's
+viewports apply their own rotations on top of the world axes: the 2-D plan
+panel wraps its content in a 90° rotation (see `FLOORPLAN_VIEW_ROTATION_DEG`
+in `packages/editor/src/components/editor/floorplan-panel.tsx`), and the 3-D
+"top-down" snap defaults to a ~45° offset between world and screen axes when
+you snap to it from the iso default position. So a layout authored as if
+*"Y = north, viewed top-down"* — common in land surveys, north-up site plans,
+and 2-D plotting libraries — will arrive **rotated** relative to its source
+when viewed in Pascal (and possibly further reflected, depending on which
+viewport and camera state you're in). The editor's own 2-D and 3-D tools are
+internally consistent with their stored coordinates, so this only affects
+geometry authored programmatically. To verify orientation before trusting
+externally-computed coordinates, place a scaled guide image at known anchor
+points and check alignment; apply whatever rotation (or reflection) your
+authoring side needs to match.
+
+A worked demonstration of all of this — axis-aligned baseline, the rotated
+30° example below, and a paired "page-intent vs world-result" L for the
+external-coordinate gotcha — lives in
+[`examples/coordinate-conventions-demo.md`](./examples/coordinate-conventions-demo.md)
+and [`examples/coordinate-conventions-demo.json`](./examples/coordinate-conventions-demo.json).
+Load the JSON with `pascal-mcp --scene examples/coordinate-conventions-demo.json`
+or import it through any host configured for the package.
 
 **Example — a 6 × 4 m slab rotated 30° about its first corner** (not
 axis-aligned, so the mapping is actually exercised):
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -227,6 +227,55 @@ console.log(scene)
 See [`examples/embed-in-agent.ts`](./examples/embed-in-agent.ts) for a
 compilable version.
 
+## Coordinate conventions
+
+Pascal is a **right-handed** scene where **X and Z form the ground plane and Y
+is up**. Lengths are in **metres**; rotations are **radians**, stored as Euler
+`[x, y, z]` tuples.
+
+**Plan → world.** Every 2-D point you pass is a ground-plane coordinate
+`[x, z]` — this includes `wall.start` / `wall.end` and the `polygon` / `holes`
+arrays of `slab`, `zone`, and `ceiling`. It maps to world space as:
+
+```
+[x, z]  →  (x, y, z)      // the 2nd component is world Z (depth), not "up"
+```
+
+There is no sign flip. The editor builds plan geometry as
+`new Vector3(point[0], y, point[1])` (see
+`packages/nodes/src/{wall,slab,ceiling}/tool.tsx`). The vertical `y` comes from
+the owning `level`'s elevation (its `level` field, in metres) plus the
+element's own height; slabs additionally carry an absolute `elevation`.
+
+**Heads-up when you compute coordinates outside the editor.** Because the plan
+is the X–Z plane, the second coordinate is **depth**, not a screen-"up" axis.
+If you generate geometry from a source that assumes *"Y = north, viewed from
+the top"* — a land survey with bearings from North, a north-up site plan, or
+most 2-D plotting libraries — the result can arrive **mirrored** relative to
+that source when viewed top-down. Pascal's own 2-D and 3-D tools are
+internally consistent, so this only affects geometry authored programmatically.
+Decide your axis mapping explicitly and verify orientation against a known
+reference (e.g. a scaled guide image) before trusting it; reflect across the
+appropriate axis if it comes in flipped.
+
+**Example — a 6 × 4 m slab rotated 30° about its first corner** (not
+axis-aligned, so the mapping is actually exercised):
+
+```json
+{
+  "op": "create",
+  "parentId": "<levelId>",
+  "node": {
+    "type": "slab",
+    "elevation": 0.0,
+    "polygon": [[0, 0], [5.196, 3.0], [3.196, 6.464], [-2.0, 3.464]]
+  }
+}
+```
+
+This lands flat on the ground (Y = 0), 6 m along a heading 30° off the +X axis
+and 4 m along its perpendicular — i.e. occupying world (x, z) directly.
+
 ## Tools
 
 All tools validate their inputs and outputs with Zod. Mutation tools are
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -233,26 +233,28 @@ Pascal is a **right-handed** scene where **X and Z form the ground plane and Y
 is up**. Lengths are in **metres**; rotations are **radians**, stored as Euler
 `[x, y, z]` tuples.
 
-**Plan → world.** Every 2-D point you pass is a ground-plane coordinate
+**Plan → world.** Every 2-D point you pass is a level/building-local
+ground-plane coordinate
 `[x, z]` — this includes `wall.start` / `wall.end` and the `polygon` / `holes`
-arrays of `slab`, `zone`, and `ceiling`. It maps to world space as:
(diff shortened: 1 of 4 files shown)


Aymericr (dismissed):
Thanks for putting this together — really thoughtful contribution, and the issue you opened was right that this convention belonged in the README. The 30° worked example, the rotate-vs-mirror nuance, and the section placement before `## Tools` are all good calls.

Before we land it, a few things I'd like to tighten up:

**Blocking-ish (worth fixing):**

1. **The "world" framing is a touch overstated.** Plan coordinates are level/building-local — they only equal world `(x, y, z)` when the building transform is identity (which is the default, but worth saying explicitly so authors with rotated/translated buildings aren't surprised). Same caveat for `level.level`: it's a floor ordering value, not a metre elevation; `LevelSystem` stacks levels by computed heights.

2. **The 6×4m / 24m² claim is rounded.** With the polygon as written (`5.196`, `3.0`, `3.196`, `6.464`, `-2.0`, `3.464`) the sides come out as `5.9999, 3.9999` to 4dp and area `23.998944`. Either bump precision (e.g. `[[0,0],[5.196152,3],[3.196152,6.464102],[-2,3.464102]]`) or add a "≈" / note that the example is rounded for readability. Same in the demo md.

3. **The demo md references `coordinate-conventions-demo.py` and `.png` as "next to this file"**, but neither is committed. Either commit them (with attribution / clear provenance), or drop that reproducibility paragraph — as-is it points readers at files that don't exist.

4. **`load_scene` in the demo md doesn't do what's implied.** That tool resolves saved scenes by ID; it can't load arbitrary bundled JSON. The `pascal-mcp --stdio --scene <path>` invocation is correct, so I'd just remove the "or call `load_scene`" sentence.

5. **`packages/mcp/package.json` only ships `dist`, `README.md`, `CHANGELOG.md`** — the `examples/...` links in the README will be dead on npm unless `examples` gets added to `files`. Worth fixing in the same PR since the README now points there.

**Worth discussing (non-blocking):**

- The "top-down 45°" caveat is really *camera-state* behaviour (default iso azimuth is preserved through `handleTopView`), not a coordinate convention. It could move to its own subsection or get a one-liner clarifying that orbit buttons / fit-scene change azimuth and the offset goes away once you're axis-aligned.
- The "no sign flip" claim is correct *at the coordinate-contract level*, but the proof that "the editor builds geometry as `new Vector3(point[0], y, point[1])`" only holds for the tool files. The renderers (`slab-system`, `ceiling-system`, `wall-system`, `zone/renderer`) actually go through `Shape(x, -z)` + `rotateX(-π/2)`, which arrives at the same world result via a different path. Net behaviour is fine — but if a reader audits those files looking for the documented pattern, they'll get confused. Maybe frame it as "the stored convention is `[x, z]`; tooling consumes it as `Vector3(x, y, z)`" without pinning to one specific call site.
- One thing I think really *would* help MCP authors that's currently missing: door / window / wall-attached `place_item` use **wall-local metres** for `position[0]`, not plan coordinates (the tools even return `coordinateSystem: 'wall-local-meters'`). Wall-attached rotations are wall-local too. Worth a paragraph in the same section since it's exactly the kind of thing this PR is meant to surface.
- Source-path pins (`packages/editor/src/components/editor/floorplan-panel.tsx`, etc.) will rot. Maybe just name the constant (`FLOORPLAN_VIEW_ROTATION_DEG`) and trust readers / `rg` to find it.
- Smaller: `examples/...` links assume CWD is `packages/mcp`. From repo root the path is `packages/mcp/examples/...`. And the 2,124-line demo JSON could probably be a much smaller scene and still carry the same proof — it's a public repo and that file is most of the diff.

The core work here is solid and I'd happily land a tightened version. The math holds up, the rotate-vs-mirror correction across commits is exactly the right instinct, and the README is genuinely better with this section than without it.

Aymericr:
Pushed a small follow-up commit (4183bee7) on top of your branch with the tweaks I had in mind — rather than bouncing the PR back for them. Net diff is +29 / −26 across the same three files (plus `package.json`). Highlights:

- Reframed "plan → world" as level/building-local (only equals world under the default identity transform).
- Fixed the `level.level` description — stacked Y comes from the level system computing accumulated heights, not from `level` directly.
- Softened the `Vector3(point[0], y, point[1])` pin since the renderers (`slab-system`, `ceiling-system`, etc.) actually go through `Shape(x, -z)` + `rotateX(-π/2)` and arrive at the same world result via a different path. Wording now describes the stored contract rather than one specific call site.
- Reworded the top-down 45° caveat as azimuth inheritance (the snap preserves azimuth; the iso default is what produces the offset).
- Marked the 6×4 m example as rounded ("sides ≈ 6 m / 4 m at 3 dp") so the polygon as-written isn't claimed to be exact.
- Added a short note that door/window `position[0]` and wall-attached `place_item` use wall-local metres, not plan coords — felt like the same kind of trap the section is meant to surface.
- Dropped the `load_scene` instruction in the demo md (that tool loads saved scenes by id, can't import bundled JSON) and the missing-script "Reproducibility" paragraph.
- Added `examples` to `package.json#files` so the README's `./examples/...` links don't 404 on npm.

Holler if any of those rewordings landed wrong — happy to adjust before merging.

PR #435, opened 2026-06-22
fix(editor): set GLTFExporter textureUtils for compressed-texture GLB export (Sentry EDITOR-DJ)

anton-pascal:
## Sentry: MONOREPO-EDITOR-DJ (production)

```
THREE.GLTFExporter: setTextureUtils() must be called to process compressed textures.
```

### Root cause
GLB export (`GLTFExporter`) of a scene containing compressed (KTX2/basis) textures throws under three.js r184. r184's `GLTFExporter` requires a `textureUtils` object (backed by the active `WebGLRenderer`) to decompress compressed textures during export. Without it, `exporter.parse(..., { binary: true })` throws and the export fails.

### Fix
In `packages/editor/src/components/editor/export-manager.tsx`:
- Pull the active renderer via `const gl = useThree((state) => state.gl)`.
- Import the decompress helper: `import * as WebGLTextureUtils from 'three/examples/jsm/utils/WebGLTextureUtils.js'`.
- Call `exporter.setTextureUtils({ decompress: (texture, maxTextureSize) => WebGLTextureUtils.decompress(texture, maxTextureSize, gl) })` before `exporter.parse(...)`.
- Added `gl` to the `useEffect` dependency array.

This lets the exporter decompress KTX2/compressed textures using the live renderer, resolving the throw.

### Verification
- `bun run check-types` (package `@pascal-app/editor`): **no errors introduced in `export-manager.tsx`**. (Pre-existing, unrelated workspace dependency-version typecheck errors in other files remain and are out of scope for this fix.)
- `biome check` on the touched file: clean, no fixes applied.

> Do not auto-merge — opening for review only.

---

> [!NOTE]
> **Low Risk**
> Scoped to GLB export in `export-manager.tsx` with an isolated decompress path that avoids touching the live WebGPU canvas renderer.
> 
> **Overview**
> Fixes production GLB export failures when the scene uses **KTX2/basis compressed textures**, which three.js r184’s `GLTFExporter` refuses to process unless `setTextureUtils()` is configured.
> 
> Before `exporter.parse`, the export path now registers **`WebGPUTextureUtils.decompress`** (matching the editor’s **WebGPU** renderer) via `exporter.setTextureUtils({ decompress: ... })`. The live canvas renderer is **not** passed into `decompress`, because that helper can resize the renderer and leave the visible viewport corrupted; omitting it lets three use a disposable renderer for the blit.
> 
> STL/OBJ export behavior is unchanged; only the default GLB branch gains this hook.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 986995fb8223808a85d326550cb20fc7522a11ad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/editor/src/components/editor/export-manager.tsx
+++ b/packages/editor/src/components/editor/export-manager.tsx
@@ -7,6 +7,7 @@ import type { Mesh, Object3D } from 'three'
 import { GLTFExporter } from 'three/examples/jsm/exporters/GLTFExporter.js'
 import { OBJExporter } from 'three/examples/jsm/exporters/OBJExporter.js'
 import { STLExporter } from 'three/examples/jsm/exporters/STLExporter.js'
+import * as WebGPUTextureUtils from 'three/examples/jsm/utils/WebGPUTextureUtils.js'
 
 export function ExportManager() {
   const scene = useThree((state) => state.scene)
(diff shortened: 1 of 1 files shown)


Aymericr:
Reworked from a dual review pass. Two changes vs the original:
- **Use `WebGPUTextureUtils` instead of `WebGLTextureUtils`:** the app renders with `THREE.WebGPURenderer`, so the WebGPU texture utils are the correct match for r184.
- **Don't pass the live renderer to `decompress()`:** both utils call `renderer.setSize(w,h)` and never restore it, which would corrupt the visible canvas after export. Omitting the renderer lets three spin up and dispose its own throwaway renderer for the blit.

Typecheck passes. Verifying CI before merge.

PR #480, opened 2026-07-09
fix(items): retry failed model loads, settle skipped items, keep exports clean

wass08:
## What does this PR do?

Makes item model loading resilient for both interactive use and headless bakes. Today a transient storage failure (observed in prod: Supabase 504s under the bake page's ~200-concurrent-request asset burst) permanently breaks an item for the session — drei's `useGLTF` caches the rejection — and worse, the red debug-box fallback gets **baked into the exported GLB** (observed in a prod artifact: 24-vert boxes with the red translucent material where 4 kettles should be).

- **Bounded retries** (1s/3s): clear the `useGLTF` cache entry and re-mount via a new `resetKey` on `ErrorBoundary`. The retry timer is effect-owned, keyed on failure count — StrictMode's synthetic unmount/remount re-arms it (an onError-scheduled timer dies silently in dev, verified).
- **Exhausted retries → item settles as skipped**: the debug box renders nothing while `isExporting`, and the failure is recorded in `useViewer.itemLoadFailures` (nodeId → url) so a bake host can persist which items are missing.
- **Scene-ready waits for item content**: `ItemSystem` now holds the dirty mark until the item settles (model mounted / terminally failed / never expected) instead of clearing at group registration — closing a race where a bake could export loading placeholders; the loading placeholder also hides during exports.

## How to test

1. `bun dev`, open a scene with items — loading placeholders, hover/selection, and slot materials behave as before.
2. Simulate a dead asset (devtools → block an item's model URL, reload): after ~5s (2 retries) the item shows the red debug box; the scene stays interactive.
3. Bake locally with the URL blocked: the exported GLB contains no debug box for that item and the bake completes; unblock, re-bake: item present.

Verified against a 200+-item prod scene: permanent 504 → exactly 3 fetch attempts then a clean skip; 504-once → retry heals, artifact byte-equivalent to intact; no-failure runs byte-identical (demo_1).

N/A — failure-path behavior; exported-GLB assertions above are the evidence.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes item load lifecycle and scene-ready gating used by headless GLB bakes; incorrect settlement timing could delay exports or omit items, but happy-path behavior is intended to stay byte-identical.
> 
> **Overview**
> Item GLB loading is now resilient to transient CDN/storage failures and headless bakes no longer capture debug or loading placeholders.
> 
> **`ModelWithRetry`** wraps item loads with bounded retries (1s / 3s): on failure it clears drei's `useGLTF` cache, bumps an `ErrorBoundary` `resetKey`, and re-mounts. After retries are exhausted the item **settles as skipped**—the red debug box stays in dev but renders **nothing** when `isExporting`; failures are recorded in **`useViewer.itemLoadFailures`** (nodeId → URL) for bake metadata.
> 
> **Scene-ready / bake timing:** `ItemRenderer` sets `userData.itemModelSettled` when the model mounts, fails terminally, or isn't needed (e.g. room-clear preview). **`ItemSystem`** only clears the dirty mark once settled, so scene-ready and exports wait for real geometry instead of grey loading boxes.
> 
> **Export hygiene:** `PreviewModel` and `BrokenItemFallback` return `null` during export. **`ErrorBoundary`** gains `onError` and `resetKey` to support the retry flow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2b032cc2dafc73f8a88ee4b01f130242b2976a45. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/nodes/src/item/renderer.tsx
+++ b/packages/nodes/src/item/renderer.tsx
@@ -36,7 +36,7 @@ import { useAnimations } from '@react-three/drei'
 import { Clone } from '@react-three/drei/core/Clone'
 import { useGLTF } from '@react-three/drei/core/Gltf'
 import { useFrame } from '@react-three/fiber'
-import { Suspense, useEffect, useLayoutEffect, useMemo, useRef } from 'react'
+import { Suspense, useCallback, useEffect, useLayoutEffect, useMemo, useRef, useState } from 'react'
 import type { AnimationAction, Group, Material, Mesh } from 'three'
 import { MathUtils } from 'three'
 import { positionLocal, smoothstep, time } from 'three/tsl'
@@ -181,6 +181,7 @@ const resolveItemMaterial = (
 const BrokenItemFallback = ({ node }: { node: ItemNode }) => {
   const handlers = useNodeEvents(node, 'item')
   const shading = useViewer((s) => s.shading)
+  const isExporting = useViewer((s) => s.isExporting)
   const [w, h, d] = node.asset.dimensions
   const material = useMemo(() => {
     const next = createDefaultMaterial('#ef4444', 1, shading) as MutableMaterial
@@ -191,6 +192,10 @@ const BrokenItemFallback = ({ node }: { node: ItemNode }) => {
     return next
   }, [shading])
 
+  // Debug affordance only — a bake must never ship the red placeholder box
+  // (observed baked into a prod artifact when an item GLB 504'd mid-capture).
+  if (isExporting) return null
+
   return (
     <mesh position-y={h / 2} {...handlers}>
       <boxGeometry args={[w, h, d]} />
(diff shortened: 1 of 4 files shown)


wass08:
Addressed both Bugbot findings:

- **Stale settled flag on URL change** — fixed in the follow-up commit: `ModelWithRetry` is now keyed by `node.asset.src` (fresh retry budget per URL) and un-settles the item on mount, so a model swap is loaded and awaited like a first load. Re-validated the 504 scenarios (permanent → 3 attempts + clean skip; transient → heals).
- **Scene ready beats item settlement** — the 180-frame cap firing with unsettled items is the pre-existing backstop semantics (better than hanging forever; before this PR the same capped export shipped placeholder *boxes*, now it ships nothing for those items). The silent part is being closed on the consumer side: the bake page will report any still-dirty item nodes at export time into the bake metadata alongside `itemLoadFailures` (pascalorg/private-editor#198), so a cap-forced incomplete artifact is always queryable.

PR #485, opened 2026-07-10
fix(items): don't clobber a same-commit settled flag (cached-model race)

wass08:
## What does this PR do?

Fixes the effect-ordering race behind tonight's slow office bake: a **cached** item model mounts in the same commit as `ModelWithRetry`, child effects run first (`settled=true`), and the wrapper's mount effect then stomped it back to `false` permanently. The item renders and exports fine — the artifact was complete — but its dirty mark never clears, so scene-ready waits out the full `sceneReadyMaxWaitMs` cap and the bake page falsely reports the loaded items as skipped.

Only the bake worker hits this (shared warm HTTP context → synchronous cached mounts); every fresh-browser local test loads via suspense and settles after the wrapper's effect, which is why it validated clean locally.

The settled flag is now URL-stamped: `unsettleForUrl` is a no-op when the flag is already settled **for the same URL** (the cached same-commit case), and still resets on a genuine model swap (the URL-keyed remount keeps the retry-budget reset).

## How to test

1. `bun dev`, `/bake/demo_office?disable=postFx,draw` — settles organically (~2.5s), full artifact, `skippedItems` absent.
2. Real proof post-merge: a warm office-scene bake on the worker should settle in seconds instead of waiting the full 240s cap, with no false skips.

N/A.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches scene-ready / bake gating via `itemModelSettled` in ItemSystem; behavior change is narrow but timing-sensitive for headless exports.
> 
> **Overview**
> Fixes a React effect-ordering bug where **cached** GLBs could render correctly but never clear the item “settled” dirty mark, so headless bakes waited until `sceneReadyMaxWaitMs` and falsely reported skips.
> 
> **`ItemRenderer`** replaces the boolean `setSettled` callback with a **`SettleControl`** that writes `itemModelSettled` together with **`itemModelSettledUrl`**. **`ModelWithRetry`** calls **`unsettleForUrl(url)`** on mount instead of always forcing unsettled — it is a no-op when the flag is already settled for that same URL (child `ModelRenderer` effect runs first on synchronous cache hits). Genuine asset swaps still reset because the component is keyed by URL.
> 
> **`settle(url)`** is used when the model resolves, on terminal load failure, and for room-clear preview.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab3db67ff265f98ef1cd19ce28b44b8ccece2356. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/nodes/src/item/renderer.tsx
+++ b/packages/nodes/src/item/renderer.tsx
@@ -217,10 +217,10 @@ const MODEL_RETRY_DELAYS_MS = [1_000, 3_000]
  */
 const ModelWithRetry = ({
   node,
-  setSettled,
+  settleControl,
 }: {
   node: ItemNode
-  setSettled: (value: boolean) => void
+  settleControl: SettleControl
 }) => {
   // `failures` counts boundary catches; `epoch` bumps after each cache clear
   // to reset the boundary and re-mount the loader. The retry timer is owned by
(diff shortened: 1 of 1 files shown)


Aymericr:
Thanks for isolating the cached-model settle race. Current main now fixes the ordering directly: ModelWithRetry clears itemModelSettled in a parent useLayoutEffect before ModelRenderer's passive completion effect can mark the cached model settled. That landed in #505 (ae87ca547), so it avoids both the original same-commit clobber and the stale same-URL early-return identified by Bugbot here.

Because this branch conflicts with the current loader/retry implementation and its remaining change is superseded, I’m closing it. The report was useful in validating the current ordering contract.

PR #488, opened 2026-07-11
feat: add Fireplace node with animated procedural fire

toolucid:
## What does this PR do?

Adds a new architectural Fireplace node kind to the editor — the first node with an animated procedural effect (instanced fire particles + flickering point light via `useFrame`).

- 4 fireplace styles: `wall`, `freestanding`, `corner`, `double-sided`
- Parametric mantel, hearth, surround, and firebox with inspector controls
- Animated instanced fire particle system with additive blending + a flickering point light
- 5 fire intensity levels (`none` → `roaring`), 4 fire colors (`orange`, `amber`, `blue`, `white`)
- Placement tool with grid snap, alignment guides, SFX, and translucent ghost preview
- Resize / rotate / move handles in the 3D viewport

Registered through the builtin plugin via the same `NodeDefinition` contract as all other node kinds. Follows the three-checkbox model from `wiki/architecture/node-definitions.md`: `geometry` (pure builder) + `renderer` (custom React with `useFrame`) + `tool` (lazy placement component).

## How to test

1. `git clone https://github.com/toolucid/editor.git && cd editor && bun install && bun dev`
2. Open `http://localhost:3002` in a browser
3. In the left sidebar palette, find the Fireplace under the Furnish section
4. Click it, then click on the floor to place a fireplace
5. Verify: mantel, hearth, surround, and firebox render correctly; fire particles animate upward with a flickering light
6. Select the fireplace and use the inspector to change fire intensity (`none` → `roaring`), fire color, and fireplace style
7. Verify resize / rotate / move handles work in the 3D viewport
8. Run `bun check-types` and `bun check` — both pass with 0 errors

Will add a screen recording before review.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Adds a new discriminated union member and event surface area, plus per-instance `useFrame` animation and extra lights that could affect viewer performance with many fireplaces; otherwise mirrors existing furnish nodes with no auth or persistence changes.
> 
> **Overview**
> Introduces a new **`fireplace`** furnish node end-to-end: Zod schema in core, `FireplaceEvent` on the editor bus, and registration in the builtin plugin.
> 
> The node supports parametric dimensions (mantel, hearth, surround, firebox), four **styles** (`wall`, `freestanding`, `corner`, `double-sided`), and inspector **fire** settings (intensity and color). Implementation follows the standard **`NodeDefinition`** split: box-based geometry builder, a **custom renderer** with instanced particles and a flickering point light driven by `useFrame`, floor placement **tool** (grid snap, alignment guides, ghost preview), and viewport resize/rotate/move **handles**.
> 
> Also updates `apps/ifc-converter/next-env.d.ts` to import Next route types from `./.next/types/routes.d.ts` instead of the dev path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8d3b372bf82de6d2992529fdb465c908d8127b2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/ifc-converter/next-env.d.ts
+++ b/apps/ifc-converter/next-env.d.ts
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/dev/types/routes.d.ts";
+import "./.next/types/routes.d.ts";
 
 // NOTE: This file should not be edited
 // see https://nextjs.org/docs/app/api-reference/config/typescript for more information.
--- a/packages/core/src/events/bus.ts
+++ b/packages/core/src/events/bus.ts
@@ -17,6 +17,7 @@ import type {
   ElevatorNode,
   EyebrowVentNode,
   FenceNode,
+  FireplaceNode,
   GuideNode,
   GutterNode,
   HvacEquipmentNode,
@@ -86,6 +87,7 @@ export interface NodeEvent<T extends AnyNode = AnyNode> {
 
 export type WallEvent = NodeEvent<WallNode>
 export type FenceEvent = NodeEvent<FenceNode>
+export type FireplaceEvent = NodeEvent<FireplaceNode>
 export type ItemEvent = NodeEvent<ItemNode>
 export type SiteEvent = NodeEvent<SiteNode>
 export type BuildingEvent = NodeEvent<BuildingNode>
@@ -254,6 +256,7 @@ type RoomPresetEvents = {
 type EditorEvents = GridEvents &
   NodeEvents<'wall', WallEvent> &
   NodeEvents<'fence', FenceEvent> &
+  NodeEvents<'fireplace', FireplaceEvent> &
   NodeEvents<'item', ItemEvent> &
   NodeEvents<'site', SiteEvent> &
   NodeEvents<'building', BuildingEvent> &
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -11,6 +11,7 @@ export type {
   ElevatorEvent,
   EventSuffix,
   FenceEvent,
+  FireplaceEvent,
   GridEvent,
   GuideEvent,
   GutterEvent,
--- a/packages/core/src/schema/index.ts
+++ b/packages/core/src/schema/index.ts
@@ -70,6 +70,12 @@ export {
 } from './nodes/elevator'
 export { EyebrowVentNode } from './nodes/eyebrow-vent'
 export { FenceBaseStyle, FenceNode, FenceStyle } from './nodes/fence'
+export {
+  FireplaceMaterialRole,
+  FireplaceNode,
+  FireplaceStyle,
+  FireStyle,
+} from './nodes/fireplace'
 export { GuideNode, GuideScaleReference } from './nodes/guide'
 export { GutterNode, GutterOutlet } from './nodes/gutter'
 export { HvacEquipmentNode } from './nodes/hvac-equipment'
--- a/packages/core/src/schema/nodes/fireplace.ts
+++ b/packages/core/src/schema/nodes/fireplace.ts
@@ -1,0 +1,69 @@
+import dedent from 'dedent'
+import { z } from 'zod'
+import { BaseNode, nodeType, objectId } from '../base'
+import { MaterialSchema } from '../material'
+
+export const FireplaceMaterialRole = z.enum(['mantel', 'surround', 'hearth', 'firebox'])
+export type FireplaceMaterialRole = z.infer<typeof FireplaceMaterialRole>
+
+export const FireplaceStyle = z.enum(['wall', 'freestanding', 'corner', 'double-sided'])
+export type FireplaceStyle = z.infer<typeof FireplaceStyle>
+
+export const FireStyle = z.enum(['none', 'small', 'medium', 'large', 'roaring'])
+export type FireStyle = z.infer<typeof FireStyle>
+
+export const FireplaceNode = BaseNode.extend({
+  id: objectId('fireplace'),
+  type: nodeType('fireplace'),
+
+  material: MaterialSchema.optional(),
+  materialPreset: z.string().optional(),
+  mantelMaterial: MaterialSchema.optional(),
+  mantelMaterialPreset: z.string().optional(),
+  hearthMaterial: MaterialSchema.optional(),
+  hearthMaterialPreset: z.string().optional(),
+  fireboxMaterial: MaterialSchema.optional(),
+  fireboxMaterialPreset: z.string().optional(),
+
+  position: z.tuple([z.number(), z.number(), z.number()]).default([0, 0, 0]),
+  rotation: z.number().default(0),
+
+  style: FireplaceStyle.default('wall'),
+
+  width: z.number().min(0.6).max(4).default(1.5),
+  height: z.number().min(0.8).max(4).default(2.2),
+  depth: z.number().min(0.3).max(1.5).default(0.6),
+
+  fireboxWidth: z.number().min(0.3).max(3).default(0.9),
+  fireboxHeight: z.number().min(0.3).max(3).default(0.7),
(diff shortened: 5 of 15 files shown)


Aymericr:
This is a distinct and useful node proposal, so I’m keeping it open, but it needs a current-main revision before architectural review:

- rebase onto the current builtin-plugin/node-definition contracts and remove the unrelated generated next-env.d.ts change
- resolve the 11 inline Bugbot findings
- add schema, geometry, handle, and placement-tool tests
- define bake/export behavior for the procedural particles and light
- include the promised recording and a many-fireplace performance check; per-instance useFrame callbacks and point lights are the main runtime risk

A focused follow-up commit satisfying those points would make this reviewable as a real feature rather than a prototype snapshot.

Aymericr (changes requested):
Thanks for this — a lot of it is in the right shape. Specifically good: the node is registered in **both** required places (`AnyNode` union in `packages/core/src/schema/types.ts` and `builtinPlugin.nodes` in `packages/nodes/src/index.ts`), `buildFireplaceGeometry` is a genuinely pure `(node) => Group` builder, the tool reuses the shared helpers in `packages/nodes/src/shared/floor-placement.ts` instead of re-rolling raycasts, and the `HandleDescriptor` set looks right. `bun run check` does exit 0 on this branch.

That said I can't merge it as-is — there are runtime breakages, and they mostly trace back to one architectural decision.

**The root issue: this definition declares both `geometry` and `renderer`.** `packages/nodes/src/fireplace/definition.ts:154-155` is the only definition in the repo (of 46) doing that. `<NodeRenderer>` returns early when `def.renderer` is present, so `<ParametricNodeRenderer>` never runs — but your renderer *also* calls `useRegistry` (so `<GeometrySystem>` reparents the built children onto the group) **and** mounts `<primitive object={geometry} />` from its own `useMemo` build at `renderer.tsx:200`. That's two copies of every mesh. `wiki/architecture/node-definitions.md:189` treats `geometry` and `renderer` as alternatives. Simplest fix: drop `renderer`, delete `fireplace/renderer.tsx`, and let the shared parametric path own the mount. Almost everything below disappears if you do.

Blocking, in order:

1. **`{...liveTransform}` NaNs the transform on drag** — `renderer.tsx:198`. `LiveTransform.rotation` is a `number`, and R3F's `applyProps` on a number without `setScalar` does `target.set(value)`. `THREE.Euler` has no `setScalar`, so `rotation.set(0.75)` leaves y/z `undefined` and `updateMatrix()` yields an all-NaN matrix — I ran this to confirm. The fireplace vanishes mid-drag. `parametric-node-renderer.tsx:67-82` normalizes rotation to a tuple exactly to avoid this.
2. **Degrees/radians mismatch** — `renderer.tsx:195` does `(node.rotation * Math.PI) / 180`, but `packages/core/src/schema/nodes/fireplace.ts:29` `rotation` is radians like every other node. Should be `rotation={[0, node.rotation, 0]}`. (`cornerAngle` *is* degrees and does need the conversion — that one's fine.)
3. **`markDirty` in the render body** — `renderer.tsx:184` writes to the store during render, re-triggering a rebuild each commit. Move it into a `useLayoutEffect` keyed on the geometry params. Related: #556 is fixing the same class of thing for walls.
4. **Corner style's yaw is silently dropped** — `geometry.ts:176` sets `group.rotation.y`, but `<GeometrySystem>` reparents only `built.children` and never applies the built group's transform. Wrap the 8 meshes in an inner rotated `Group` (returned as the root's single child), or bake the rotation into the children.
5. **`mantelHeight` is a dead control** — `geometry.ts:164` only uses it as a `> 0` gate; it never affects the mantel's size or Y. I diffed 0.05 vs 0.50 and got byte-identical `BoxGeometry` params and the same position, yet it's exposed in `parametrics.ts:26`. Either wire it to the thickness or remove the control.
6. **`capabilities.floorPlaced` is missing** — `definition.ts:141-150`. Without it `getFloorPlacedElevation` early-returns 0 (`packages/core/src/hooks/spatial-grid/floor-placed-elevation.ts:67-75`), so the `getFloorStackPreviewPosition` call in `tool.tsx:77-82` is inert and a fireplace on a raised slab sinks through it. Also add `'fireplace'` to `FLOOR_PLACEMENT_CLICK_TRIGGER_KINDS` (`shared/floor-placement.ts:16-28`) so the placement click actually fires.

Should-fix:

- **No disposal anywhere** — the `useMemo`'d group, the ember `SphereGeometry`/`MeshBasicMaterial`, and the built geometry in `preview.tsx` all leak. `cupola/renderer.tsx` has the pattern (`useEffect(() => () => geometry.dispose(), [geometry])`).
- **Raw per-node `<pointLight>` + two `useFrame` callbacks** (`renderer.tsx:127, 75, 120`) bypass the pooled light system (`packages/viewer/src/systems/item-light/item-light-system.tsx`, POOL_SIZE = 12, with camera scoring and hysteresis). Register a `LightEffect` via `useItemLightPool` — see `packages/nodes/src/item/renderer.tsx:805`. Twenty fireplaces currently means 20 lights and 40 frame callbacks.
- **No tests.** A `fireplace/geometry.test.ts` covering child count per style, `mantelHeight` affecting dimensions, corner yaw landing on children, and the lintel staying inside `node.height` at the schema minimum (0.8 currently puts the lintel top at 1.17 m) would lock all of the above in.
- CI has never run on `a8d3b37` — please push so the required checks execute.

Nits: `renderer.tsx:152` uses `// eslint-disable-next-line react-hooks/exhaustive-deps`, which is inert here — this repo uses `// biome-ignore lint/correctness/useExhaustiveDependencies:`. The `apps/ifc-converter/next-env.d.ts` change is unrelated to the feature and reverts a generated path `main` still carries; please drop it. The four near-identical material resolvers in `geometry.ts` could collapse into one `resolveMaterial(node, role)`, which would also give the exported `FireplaceMaterialRole` a use. And there's no `floorplan` builder, no `mcp` block (this would be the only definition of 46 without one), and no `bake` policy — worth deciding deliberately whether the fireplace should show in 2D plans, be MCP-addressable, and survive GLB export.

Two things that are **not** on you: the `packages/core/src/events/bus.ts` conflict and the removed `packages/plugin-trees` are just base drift (82 commits since merge base `7d6879c4`) — a rebase clears both. And the local bun-test failures reproduce identically on `main`, so they're environmental.

One process question before you invest in the fixes: I couldn't find an issue requesting a fireplace node. Could you say a bit about the use case? If it's wanted, the geometry builder here is a good starting point and the fix list gets much shorter once the custom renderer is dropped.

Aymericr:
Closing this as stale — seven weeks with no reply to either review round, and main has moved 202 commits underneath it.

Some of that movement is worth knowing if you come back to it. The `renderer` field is now a tagged union (`{ kind: 'parametric', module: () => import('./renderer') }`), so `definition.ts` no longer typechecks as written. `FLOOR_PLACEMENT_CLICK_TRIGGER_KINDS` is gone from `shared/floor-placement.ts`. And my "only definition doing both `geometry` and `renderer`" line has aged out — `lean-to-extension` now does exactly that, and the wiki blesses it. What hasn't changed is the bug behind it: this renderer calls `useRegistry` *and* mounts its own `<primitive>`, so every mesh draws twice. `lean-to-extension` is the pattern to copy if you want to keep the custom renderer.

The rest of the 4 August list is still open too — the `{...liveTransform}` rotation NaN, degrees vs radians, `markDirty` in the render body, the corner yaw never reaching the children, `mantelHeight` as a dead control, no disposal, and per-node `<pointLight>` instead of `useItemLightPool` (the pool is 12 lights with camera scoring for precisely this reason).

`buildFireplaceGeometry` is genuinely good — a clean pure builder, and the tool reuses the shared floor-placement helpers instead of re-rolling raycasts. If a fireplace is something you want in your own scenes, open an issue describing the use case and I'll happily help scope a smaller version: geometry-only, no custom renderer, fire as a later follow-up. That would be a much shorter path to merged than this branch is now.

Thanks for building it, and no hard feelings in the close — reopen any time.

PR #530, opened 2026-07-22
fix(nodes): guard missing roof-segment trim in ridge-vent geometry key (Sentry MONOREPO-EDITOR-ED)

anton-pascal:
## Summary
- tolerate roof segments whose stored data predates the `trim` schema field
- keep the ridge-vent geometry dependency key stable by reading trim fields from an empty partial object when `segment.trim` is missing

## Root cause
The public viewer hydrates raw stored nodes without re-running the Zod parser. For legacy roof segments, that means the `RoofSegmentTrim` schema default never applies and `segment.trim` can be `undefined` at runtime. `ridgeVentSegmentGeometryKey` then dereferenced `trim.left`, crashing the viewer.

`buildRidgeVentGeometry` already routes trim handling through `normalizeRoofSegmentTrim`, so no additional guard was needed there.

## Sentry
- [MONOREPO-EDITOR-ED](https://pascalorg.sentry.io/issues/7593979332/)
- Issue `7593979332`
- Observed July 22, 2026 on `/viewer/project_myTyG1YFgGBVpS6O`

## Validation
- `bun test packages/nodes/src/ridge-vent/__tests__/geometry.test.ts` — 18 pass, 0 fail
- `bunx biome check packages/nodes/src/ridge-vent/renderer.tsx` — clean

---

> [!NOTE]
> **Low Risk**
> Small defensive change in ridge-vent rendering; aligns geometry key with existing normalization used elsewhere.
> 
> **Overview**
> Fixes a **viewer crash** when ridge vents render on **legacy roof segments** that were saved before the `trim` field existed. The public viewer hydrates nodes without re-running Zod, so `segment.trim` can be `undefined` even though parsed segments get defaults.
> 
> **`ridgeVentSegmentGeometryKey`** now builds its cache key from **`normalizeRoofSegmentTrim(segment)`** instead of reading `segment.trim` directly, matching how **`buildRidgeVentGeometry`** already handles trim. A core test documents that segments with no `trim` at all normalize to zero trim.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1cdf09a3c419b95ea7119783703875db19f3d7cc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/schema/nodes/roof-segment-trim.test.ts
+++ b/packages/core/src/schema/nodes/roof-segment-trim.test.ts
@@ -32,6 +32,15 @@ describe('roof segment trim', () => {
     })
   })
 
+  // Scene loading casts migrated roof-segments instead of zod-parsing them, so a
+  // segment saved before `trim` existed reaches renderers with the field absent.
+  test('tolerates a segment carrying no trim at all', () => {
+    const trim = normalizeRoofSegmentTrim({ width: 8, depth: 6 })
+
+    expect(trim.left).toBe(0)
+    expect(trim.backRightZ).toBe(0)
+  })
+
   test('normalizes impossible side totals without inverting the footprint', () => {
     const trim = normalizeRoofSegmentTrim({
       width: 4,
--- a/packages/nodes/src/ridge-vent/renderer.tsx
+++ b/packages/nodes/src/ridge-vent/renderer.tsx
@@ -4,6 +4,7 @@ import {
   type AnyNodeId,
   getEffectiveRoofSurfaceMaterial,
   getEffectiveSegmentSurfaceMaterial,
+  normalizeRoofSegmentTrim,
   type RidgeVentNode,
   type RoofNode,
   type RoofSegmentNode,
@@ -29,7 +30,7 @@ import { buildRidgeVentGeometry } from './geometry'
 
 function ridgeVentSegmentGeometryKey(segment: RoofSegmentNode | undefined): string {
   if (!segment) return 'none'
-  const trim = segment.trim
+  const trim = normalizeRoofSegmentTrim(segment)
   return [
     segment.roofType,
     segment.width,


Aymericr:
Confirmed the crash is reachable and taken this slightly further than the original patch.

**Why the null is real**, since it is worth recording: `trim` is a required field with a schema default, so it looks unreachable. But the roof-segment branch of the scene-load migration in `use-scene.ts` builds its result with a cast rather than a zod parse — the comment there says so explicitly for `pitch`. That means a segment saved before `trim` existed arrives at the renderer with the field genuinely absent, and the schema default never lands. The Sentry report is not a phantom.

**What I changed.** The patch used `segment.trim ?? {}` with a `Partial<RoofSegmentTrim>` annotation. That fixes the crash, but the two sibling call sites — `shared/use-segment-trim-clip.tsx` and `ridge-vent/geometry.ts` — both route through core's `normalizeRoofSegmentTrim`, so I switched to that for consistency.

It also fixes a second, quieter problem. `normalizeRoofSegmentTrim` clamps impossible trims: a `left`/`right` pair that would invert the footprint gets scaled to fit. The geometry this key guards is built from the *normalized* trim. Keying on the raw field means two distinct raw values that normalize to the same clamped result produce different keys, so the `useMemo` rebuilds geometry that is byte-identical. Reading through the same normalizer makes the cache key agree with what is actually rendered.

Added a regression test in `roof-segment-trim.test.ts` for the trim-absent case, with a comment recording *why* it is reachable so nobody deletes it as redundant with the schema default.

```
bun test roof-segment-trim ridge-vent  → 25 pass, 0 fail
bun run check                          → 1583 files, no fixes applied
```

Pushed to `fix/sentry-EDITOR-ED`. Will merge once CI is green.

PR #550, opened 2026-07-27
fix(nodes): reuse wall draft preview geometry

JimmyZheng-ZJU:
## What does this PR do?

This removes per-pointer-move geometry allocation from the 3D wall draft preview.

Previously, every `grid:move` event constructed a new `BoxGeometry`, disposed the previous geometry, and assigned the replacement to the preview mesh. Continuous wall drawing therefore created avoidable CPU allocations and GPU resource churn on the editor's hottest interaction path.

The preview now mounts one unit `boxGeometry` and updates the mesh position, rotation, and scale. Snapping, preview dimensions, visibility, and chained wall placement behavior remain unchanged.

In a local 30-event pointer-move sample, the previous path produced roughly 4 MB of short-lived heap growth. With geometry reuse, the sampled heap delta was effectively flat; exact values remain workload- and GC-dependent.

Validation performed:

- `bun test packages/nodes/src/wall` — 71 passed
- `bun run build` in `packages/nodes` — passed
- Biome check for `packages/nodes/src/wall/tool.tsx` — passed
- Manual continuous-wall checks in both the 3D editor and 2D floorplan

## How to test

1. Run `bun dev`, open a project in the 3D editor, and select the Wall tool.
2. Place the first wall segment, then move the pointer to preview and place a second connected segment; verify the preview follows continuously and retains the expected height and thickness.
3. Switch to the 2D floorplan and verify wall preview, chained placement, and cancellation still behave as before.

Not included: this is a resource-lifetime optimization with intentionally unchanged visual output. The interaction was verified manually in both 3D and 2D.

## Checklist

- [x] I've updated relevant documentation (not applicable: no public API or behavior change)

---

> [!NOTE]
> **Low Risk**
> Editor-only preview rendering optimization with unchanged snapping, dimensions, and placement behavior; visual output should match prior behavior.
> 
> **Overview**
> **Wall and straight fence draft previews** no longer allocate a new `BoxGeometry` on every `grid:move`. Both tools mount a single unit **`boxGeometry`** and drive the bar with **`mesh.scale`** plus the existing position/rotation updates in `updateWallPreview` / `updateFencePreview`, dropping per-move dispose/assign churn on the editor’s hottest path.
> 
> The preview mesh JSX switches from **`shapeGeometry`** to **`boxGeometry`**, aligned with the scale-based path and avoiding a one-frame WebGPU “vertex buffer slot 0” warning when visibility was toggled before real geometry existed. Wall-tool comments are tightened to describe hiding the unit box until the first move scales it (including after chained segments).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 57f9f9f68cbd09a46a64cffd700c3d434850ab52. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/nodes/src/fence/tool.tsx
+++ b/packages/nodes/src/fence/tool.tsx
@@ -50,15 +50,7 @@ import {
 import { getSceneTheme, useViewer } from '@pascal-app/viewer'
 import { useThree } from '@react-three/fiber'
 import { useEffect, useMemo, useRef, useState } from 'react'
-import {
-  BoxGeometry,
-  BufferGeometry,
-  type Camera,
-  DoubleSide,
-  type Group,
-  type Mesh,
-  Vector3,
-} from 'three'
+import { BufferGeometry, type Camera, DoubleSide, type Group, type Mesh, Vector3 } from 'three'
 import {
   DraftAngleArc,
   type DraftAngleLabel,
@@ -430,16 +422,11 @@ function updateFencePreview(
   }
   mesh.visible = true
   direction.normalize()
-  const geometry = new BoxGeometry(length, previewHeight, previewThickness)
   const angle = Math.atan2(direction.z, direction.x)
 
   mesh.position.set((start.x + end.x) / 2, start.y + previewHeight / 2, (start.z + end.z) / 2)
   mesh.rotation.y = -angle
-
-  if (mesh.geometry) {
-    mesh.geometry.dispose()
-  }
-  mesh.geometry = geometry
+  mesh.scale.set(length, previewHeight, previewThickness)
 }
 
 function getCurrentLevelElements(): { walls: WallNode[]; fences: FenceNode[] } {
@@ -762,7 +749,7 @@ const StraightFenceTool: React.FC = () => {
       />
       <CursorSphere height={previewHeight} ref={cursorRef} />
       <mesh layers={EDITOR_LAYER} ref={previewRef} renderOrder={1} visible={false}>
-        <shapeGeometry />
+        <boxGeometry />
         <meshBasicMaterial
           color="#ffffff"
           depthTest={false}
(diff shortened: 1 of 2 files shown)


Aymericr:
Good change — verified and extended it to the fence tool, then pushed to your branch.

**On the safety of scaling instead of rebuilding.** The thing that makes this substitution valid rather than merely equivalent-looking is that nothing raycasts against the preview mesh. I checked: there is no `intersectObject` or bounding-box read anywhere against `wallPreviewRef`, and it sits on `EDITOR_LAYER`, so it is excluded from the MRT scene pass. A non-uniform `scale` would skew normals and break a lit material, but this is a `meshBasicMaterial` with `depthTest: false`, so there is no lighting to get wrong. If either of those changed the substitution would need revisiting, which is why I am recording it here rather than in a code comment.

**What I added.** `packages/nodes/src/fence/tool.tsx` carried a byte-identical copy of `updateWallPreview` with the same per-move allocation. Fixing only the wall tool would have left the two draft tools behaving differently on the same interaction, so the fence preview now mounts a unit `boxGeometry` and scales too, and the unused `BoxGeometry` import is dropped.

**Comment rewrites.** You replaced two comments that documented WebGPU-specific failures — the "Vertex buffer slot 0 was not set" warning from drawing the empty `shapeGeometry` placeholder, and the double-paint on segment commit. Trimming them was the right call given the placeholder is now a real `boxGeometry`: the *specific* warning that text described can no longer occur, so keeping it would have preserved a stale explanation. The behavior it guarded — visibility staying false until the first move — is still there and still commented.

```
bun run check-types            → 9 successful, 9 total
bun run check                  → 1583 files, no fixes applied
bun test packages/nodes/src    → 944 pass, 1 skip, 0 fail
```

Pushed to `fix/wall-draft-preview-allocation`. Will merge on green CI.

PR #551, opened 2026-07-27
fix(editor): use a literal dev port so `bun dev` works on Windows

evolv3ai:
## Problem

`bun dev` fails immediately on Windows. The `editor` dev task dies before Next.js starts:

```
editor:dev: $ dotenv -e ../../.env.local -- next dev --port ${PORT:-3002}
editor:dev: error: option '-p, --port <port>' argument '${PORT:-3002}' is invalid. '${PORT:-3002}' is not a non-negative number.
editor:dev: error: script "dev" exited with code 1
```

## Cause

On Windows, `bun run` executes package scripts with Bun's own built-in shell rather than handing them to a POSIX shell. That shell does not implement default-value parameter expansion, so `${PORT:-3002}` in `apps/editor/package.json` reaches Next.js verbatim and is rejected.

Setting `PORT` in the environment first does **not** work around it — the literal is never expanded either way. There is no shell-level workaround from the caller's side; the script itself has to avoid the syntax.

## Fix

Use the documented default port directly:

```diff
-"dev": "dotenv -e ../../.env.local -- next dev --port ${PORT:-3002}"
+"dev": "dotenv -e ../../.env.local -- next dev --port 3002"
```

Verified on Windows 11 / Bun 1.3.6 / Node 24.18.0 — `bun dev` now boots the whole workspace and the editor serves `HTTP 200` on http://localhost:3002.

## Tradeoff, and an alternative if you'd prefer it

This drops the `PORT` override that SETUP.md documents. I kept the change to one line because that's the smallest thing that unblocks Windows, but I'm happy to switch to a small cross-platform launcher instead, which would preserve `PORT` on every platform:

```js
// apps/editor/scripts/dev.mjs
import { spawn } from "node:child_process";
const port = process.env.PORT ?? "3002";
spawn("next", ["dev", "--port", port], { stdio: "inherit", shell: true });
```

Just say which you'd rather have and I'll update the PR.

## Two unrelated things I noticed while debugging

Not touched in this PR, but worth flagging:

1. **The root `dev` script is also a no-op on Windows.** It begins with `set -a && . ./.env 2>/dev/null; set +a; turbo run dev --env-mode=loose`. Bun's shell has no `set` builtin, so it prints `bun: command not found: set` twice and silently skips loading `.env`. Turbo still runs, so it's non-fatal, but any variables in a root `.env` are never loaded on Windows.

2. **`.env.example` and `SETUP.md` disagree on the default port.** `.env.example` says `# Dev server port (default: 3000)` / `# PORT=3000`, while SETUP.md and the dev script both use 3002.

---

> [!NOTE]
> **Medium Risk**
> The content-clear guard changes persistence semantics for scene PUTs (intentional clears need an explicit flag); routing and dev script changes are lower risk.
> 
> **Overview**
> **Scene save protection:** PUT `/api/scenes/[id]` now blocks writes that would remove every authored node (walls, slabs, zones, etc.) while leaving only the default site/building/level scaffold or an empty graph—matching the race where debounced autosave runs before the initial load. Those saves return **409** with `content_clear_rejected` unless the body includes **`allowContentClear: true`**. Logic lives in new `scene-content-guard` helpers with unit tests.
> 
> **Routing:** Next.js **rewrites** `/editor/:id` → `/scene/:id` so MCP/store `editorUrl` links stop 404ing without changing the browser path (client code still expects `/editor/...`).
> 
> **Dev:** `apps/editor` `dev` script uses a fixed `--port 3002` (no `${PORT:-3002}`, which breaks under Bun on Windows) and adds `--hostname 0.0.0.0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d90667b62301e34d1a1eebd5d0268decd009f3e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/apps/editor/app/api/scenes/[id]/route.ts
+++ b/apps/editor/app/api/scenes/[id]/route.ts
@@ -7,6 +7,7 @@ import {
   sceneApiPreflight,
   withSceneApiHeaders,
 } from '@/lib/scene-api-security'
+import { countContentNodes, wouldClearSceneContent } from '@/lib/scene-content-guard'
 import { getSceneOperations } from '@/lib/scene-store-server'
 
 export const dynamic = 'force-dynamic'
@@ -18,6 +19,12 @@ const putSceneSchema = z.object({
   graph: apiGraphSchema,
   thumbnailUrl: z.string().url().nullable().optional(),
   expectedVersion: z.number().int().nonnegative().optional(),
+  /**
+   * Opt out of the content-clear guard. Required for a deliberate
+   * "delete everything" save, which is otherwise indistinguishable from the
+   * client autosaving before it has loaded the scene.
+   */
+  allowContentClear: z.boolean().optional(),
 })
 
 const patchSceneSchema = z.object({
@@ -83,6 +90,30 @@ export async function PUT(request: NextRequest, { params }: RouteParams) {
     if (!existing) {
       return sceneApiJson(request, { error: 'not_found' }, { status: 404 })
     }
+
+    // A client that autosaves before its initial load resolves sends an empty
+    // graph — or the bare site/building/level scaffold — and the version still
+    // matches, so `expectedVersion` waves it straight through and the stored
+    // scene is destroyed. Refuse to be the last writer in that sequence: a save
+    // may shrink a scene freely, but it may not take the last content node with
+    // it unless the caller says so explicitly.
+    if (
+      !parsed.data.allowContentClear &&
+      wouldClearSceneContent(existing.graph, parsed.data.graph)
+    ) {
+      return sceneApiJson(
+        request,
+        {
+          error: 'content_clear_rejected',
+          details:
+            'Refusing to replace a non-empty scene with an empty one. Reload the scene, or resend with allowContentClear: true if the clear is intentional.',
+          existingContentNodes: countContentNodes(existing.graph),
+          version: existing.version,
+        },
+        { status: 409, headers: { ETag: `"${existing.version}"` } },
+      )
+    }
+
     const meta = await operations.saveScene({
       id,
       name: parsed.data.name ?? existing.name,
--- a/apps/editor/lib/scene-content-guard.test.ts
+++ b/apps/editor/lib/scene-content-guard.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, test } from 'bun:test'
+import { countContentNodes, wouldClearSceneContent } from './scene-content-guard'
+
+const scaffold = {
+  nodes: {
+    site_a: { type: 'site' },
+    building_a: { type: 'building' },
+    level_a: { type: 'level' },
+  },
+}
+
+const house = {
+  nodes: {
+    site_a: { type: 'site' },
+    building_a: { type: 'building' },
+    level_a: { type: 'level' },
+    slab_a: { type: 'slab' },
+    wall_a: { type: 'wall' },
+    wall_b: { type: 'wall' },
+    zone_a: { type: 'zone' },
+  },
+}
+
+describe('countContentNodes', () => {
+  test('ignores the site/building/level scaffold', () => {
+    expect(countContentNodes(scaffold)).toBe(0)
+  })
+
+  test('counts authored nodes only', () => {
+    expect(countContentNodes(house)).toBe(4)
+  })
+
+  test('treats an empty or malformed graph as no content', () => {
+    expect(countContentNodes({ nodes: {} })).toBe(0)
+    expect(countContentNodes({})).toBe(0)
+    expect(countContentNodes(null)).toBe(0)
+    expect(countContentNodes(undefined)).toBe(0)
+    expect(countContentNodes({ nodes: { a: null } })).toBe(0)
+  })
+})
+
+describe('wouldClearSceneContent', () => {
+  test('rejects the empty-graph autosave that wiped the stored scene', () => {
+    expect(wouldClearSceneContent(house, { nodes: {} })).toBe(true)
+  })
+
+  test('rejects a scaffold-only autosave, not just a zero-node one', () => {
+    // The regression this guards: versions 51/53/55 wrote 3-4 scaffold nodes,
+    // which a naive "is the graph empty" check would have let through.
+    expect(wouldClearSceneContent(house, scaffold)).toBe(true)
+  })
+
+  test('allows ordinary edits that shrink the scene', () => {
+    const trimmed = {
+      nodes: { site_a: { type: 'site' }, level_a: { type: 'level' }, wall_a: { type: 'wall' } },
+    }
+    expect(wouldClearSceneContent(house, trimmed)).toBe(false)
+  })
+
+  test('allows growing a scene', () => {
+    expect(wouldClearSceneContent(scaffold, house)).toBe(false)
+  })
+
+  test('allows writing a scaffold over a scene that had no content anyway', () => {
+    expect(wouldClearSceneContent(scaffold, { nodes: {} })).toBe(false)
+  })
+})
--- a/apps/editor/lib/scene-content-guard.ts
+++ b/apps/editor/lib/scene-content-guard.ts
@@ -1,0 +1,42 @@
+/**
+ * Guard against a save that silently destroys an authored scene.
+ *
+ * The editor autosaves on a debounce. If that timer fires before the initial
+ * scene load resolves, the client sends the empty graph it started from — or
+ * the bare site/building/level scaffold it falls back to. The stored version is
+ * still the one the client read, so `expectedVersion` matches and optimistic
+ * concurrency waves the write through. The authored scene is then gone.
+ *
+ * Shrinking a scene is legitimate; removing its last authored node as part of
+ * an unattended write is not distinguishable from the failure above, so callers
+ * that mean it have to say so explicitly.
+ */
+
+/**
+ * The scaffold a fresh editor session starts from. These carry no authored
+ * content on their own, so a graph containing only these is "empty" for the
+ * purposes of this guard.
+ */
+export const SCAFFOLD_NODE_TYPES: ReadonlySet<string> = new Set(['site', 'building', 'level'])
+
+/** Counts nodes a user or agent actually authored (walls, slabs, zones, openings, roofs, items). */
+export function countContentNodes(graph: unknown): number {
+  const nodes = (graph as { nodes?: Record<string, { type?: string } | null> } | null | undefined)
+    ?.nodes
+  if (!nodes || typeof nodes !== 'object') return 0
+
+  let count = 0
+  for (const node of Object.values(nodes)) {
+    const type = node?.type
+    if (typeof type === 'string' && !SCAFFOLD_NODE_TYPES.has(type)) count += 1
+  }
+  return count
+}
+
+/**
+ * True when writing `incoming` over `existing` would strip the scene of every
+ * authored node. Callers should reject such a write unless it is explicit.
+ */
+export function wouldClearSceneContent(existing: unknown, incoming: unknown): boolean {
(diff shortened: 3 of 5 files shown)


Aymericr:
Thanks — there are three real findings in here, and I verified all three. But they're three separate changes, and one of them is fixed at the wrong layer, so I'd like to split this rather than merge it as-is.

**Verified true:**

1. **The Windows port bug.** Bun Shell genuinely does not expand `${PORT:-3002}` — I tested it directly, `$\`echo ${"${PORT:-3002}"}\`` returns the literal string. On Windows, Bun runs scripts through its own shell rather than cmd/sh, so `next dev --port ${PORT:-3002}` receives a literal and fails. Real bug, and it makes the repo un-runnable on Windows out of the box.
2. **The MCP `editorUrl` 404.** `packages/mcp/src/storage/sqlite-scene-store.ts:151` returns `/editor/<id>`, and this app only has `app/scene/[id]`. Confirmed.
3. **The data-loss bug is real** — a client that autosaves before its load resolves sends the bare scaffold with a matching version, and `expectedVersion` waves it through.

**But the guard is at the wrong layer, and I found the actual root cause.**

`packages/editor/src/hooks/use-auto-save.ts` already has this guard — `isSuspiciousNodeDrop`. The reason it didn't fire is a stale counter: `lastNodeCount` is initialized at hook mount, before the scene loads, so it's the *empty* count. The `isLoadingSceneRef` branch of the subscription (line 136) refreshes `lastNodesSnapshot`, `lastCollectionsRef`, `lastMaterialsRef` and `lastInstalledPluginsRef` — but **not** `lastNodeCount`. So for the first save of every session the guard compares against a stale zero, `previousNodeCount > 4` is false, and it passes.

That matters for where the fix goes. Fixing the hook protects the hosted app and every consumer of `@pascal-app/editor`; a guard in `apps/editor`'s route protects only the OSS standalone app. And your route-level guard is bypassable one edit later: `apps/editor/components/scene-loader.tsx:131` treats any 409 as a version conflict and returns without throwing, so autosave marks the run saved and keeps the emptied graph in memory. Add one node and `wouldClearSceneContent` is false, the version still matches, and the next PUT replaces the scene with a 1-node graph. The 409 reuse also means the user sees "Conflict — reload to continue" (`save-button.tsx:91`) for what isn't a conflict — and Reload would resurrect content they may have deliberately deleted.

**What I'd like instead — three PRs:**

**A. The port fix** (please keep this one, it's the one your title describes). Rather than hardcoding, `dotenv-cli` can supply the default without any shell expansion. I verified the precedence chain:

```
"dev": "dotenv -e ../../.env.local -e ../../.env -- next dev"
```

Next's CLI already reads `PORT` from the environment (`.env('PORT')` on the `-p` option in `next/dist/bin/next`), so with `.env` carrying `PORT=3002` you get: shell `PORT` wins → else `.env.local` → else `.env`'s 3002. Works identically on Windows because nothing is expanded by a shell. That keeps the documented `PORT` override in `SETUP.md:27` and `.env.example` working — dropping it silently invalidates both, and `.env.example` is separately wrong about the default (it says 3000, it's 3002).

Also please drop `--hostname 0.0.0.0`. The comment says it pins existing behavior, but it narrows it: Next deliberately passes no default host (see the comment at `next-dev.js:194`), Node then binds `::` dual-stack, and forcing `0.0.0.0` makes `http://[::1]:3002` unreachable.

**B. The MCP URL mismatch** — worth its own PR, and I'd rather fix it in `sqlite-scene-store.ts` than paper over it with a rewrite. The rewrite's stated rationale doesn't hold here anyway: `view-toggles.tsx:136` does parse the id out of `/editor/`, but that path is dead in this repo (no `registerUploadHandler` call site).

**C. The autosave guard** — in `use-auto-save.ts`, refreshing `lastNodeCount` in the loading branch alongside the other refs. If you'd like to take that one, it's the highest-value of the three. A server-side backstop on top is defensible, but it needs its own status code rather than 409, and both clients need to handle it.

One process note: the title and body still describe a one-line dev-port fix, but the head is 5 files and +153/-1 across three commits with changed persistence semantics. That's the kind of drift that makes a PR hard to review — please update the body when scope grows. Also, the new test file lives in `apps/editor/lib/`, whose `test` script is `bun test lib`, so it does run in CI now (#548 landed the root `test` task) — worth knowing it wasn't running when you wrote it.

Happy to take A on its own immediately if you push just that.

Aymericr (changes requested):
The Windows problem is real and worth fixing — thank you for finding it. `${PORT:-3002}` is POSIX parameter expansion, `apps/editor/package.json:7` is the only place in the repo that uses it, and under `cmd.exe`/PowerShell it reaches Next as a literal string. So the diagnosis in your title is correct.

I can't merge this as it stands, for two reasons — one about the fix itself, one about scope.

**1. The fix trades one bug for another.** Next's CLI already binds `PORT` to `--port`:

```
.option('--port <port>', …).argParser(parseValidPositiveInteger).default(3000).env('PORT')
```

Commander's `.env()` is only consulted when the flag is *absent*, so hardcoding `--port 3002` makes `PORT=4000 bun dev` silently ignore `PORT` — which is how our own community app selects its port. The one-line fix that works on both platforms is to drop the flag and let Next read the env var:

```diff
-"dev": "dotenv -e ../../.env.local -- next dev --port ${PORT:-3002}",
+"dev": "dotenv -e ../../.env.local -- next dev",
```

That needs `PORT=3002` in `.env.local`/`.env.example` to preserve today's default, since Next's own default is 3000. Happy with any variant that keeps `PORT` working — `cross-env` would also do it.

Also worth knowing: this doesn't yet make root `bun dev` work on Windows. The root script is `set -a && . ./.env 2>/dev/null; set +a; turbo run dev --env-mode=loose` (`package.json:6`), which is POSIX-only too. Your change fixes `bun dev` *inside* `apps/editor`. Fixing the root script is a genuinely useful follow-up if you want it — the `.env` loading there could move to `dotenv -e ./.env` and become portable.

**2. The PR does four things and the title names one.** The diff is 153 lines across 5 files:

- `apps/editor/package.json` — the port fix (1 line), **plus an undisclosed `--hostname 0.0.0.0`**
- `apps/editor/lib/scene-content-guard.ts` + test — a new 42-line scene-persistence guard
- `apps/editor/app/api/scenes/[id]/route.ts` — 31 lines wiring that guard into the save path
- `apps/editor/next.config.ts` — an `/editor/<id>` → `/scene/<id>` rewrite

`--hostname 0.0.0.0` binds the dev server to every interface, which on a laptop on untrusted Wi-Fi exposes it to the local network. That may well be what you want in your setup, but it's a security-relevant default that needs to be its own change with its own rationale, not a rider on a port fix.

The `/editor` rewrite also collides with #570, which adds an `/editor` redirect to the same `next.config.ts`. Two PRs solving the same routing mismatch in one file will conflict, and I'd rather settle *which* behaviour is right — you rewrite (preserving the `/editor/` prefix because client code parses the id out of it), #570 redirects. Your rationale for rewriting is the more careful one, so this shouldn't be lost; it just needs to be decided in one place.

And the persistence guard is the piece I most want reviewed on its own merits: it changes what the save endpoint accepts, which is the path where a bug means data loss. It deserves a description, and reviewers looking at it deliberately — it will get neither while it's inside a PR titled "use a literal dev port".

**What I'd suggest**, and I'll review each quickly:

1. **This PR** → the one-line port fix only (using the `PORT`-preserving form above). I'll merge it as soon as it's trimmed.
2. **A second PR** → `--hostname 0.0.0.0`, with the use case. Likely answer is to make it opt-in via `HOSTNAME` rather than the default.
3. **A third PR** → the scene-content guard, with the failure it prevents.
4. **The `/editor` routing** → comment on #570 with your rewrite-vs-redirect argument and we'll land whichever is right there.

If splitting is more work than you want, tell me and I'll cherry-pick the port line myself with credit to you — but I'd genuinely rather have items 2 and 3 as reviewable changes than lose them.

Aymericr:
Two updates on my review above, one of them a correction to me.

**Finding B was my mistake — there is no 404.** I said `editorUrl` returning `/editor/<id>` was broken because this app only serves `/scene/[id]`. I missed that `apps/editor/next.config.ts` already redirects for exactly this reason:

```ts
// MCP / package metadata returns `/editor/<id>` (hosted route). This open-source
// app serves saved scenes at `/scene/<id>` — redirect so links and bookmarks work.
async redirects() {
  return [{ source: '/editor/:id', destination: '/scene/:id', permanent: false }]
}
```

And `/editor/<projectId>` is the correct route on the hosted side (`apps/community/app/editor/[projectId]`), which is what `storage/types.ts` documents the field as. So the current value is right for both consumers and the rewrite in your branch isn't needed. Sorry for sending you after that one — I went to fix it myself, got as far as a branch, and found the redirect.

**Finding C is now fixed on `main`, in the hook.** #578 landed `createStoredNodeCountTracker`, which makes the baseline refresh explicit at the point where a graph arrives from storage:

```ts
/** A graph read from storage — it defines what "populated" means from here. */
trackLoadedGraph(nodeCount: number) {
  count = nodeCount
},
```

The loading branch of the subscription now calls it alongside the other refs, so the guard is armed from the first save of the session instead of comparing against a stale zero. That's in `packages/editor`, so it covers the hosted app and every consumer of the package — not just the standalone shell. Please drop your route-level guard and the `apps/editor/lib` test when you rebase; the 409 reuse I flagged goes away with it.

**That leaves A, the fix your title actually describes.** It's still open and I'd still merge it on its own. To restate concretely:

```json
"dev": "dotenv -e ../../.env.local -e ../../.env -- next dev"
```

with `PORT=3002` in `.env`, and no `--hostname 0.0.0.0`. That keeps the documented `PORT` override working, has no shell expansion to break on Windows, and leaves dual-stack binding alone.

Push just that and I'll merge it — no need to touch the other two.

Aymericr:
All three findings are now resolved, so I'm closing this — but the Windows bug you found is fixed on `main`, and it's fixed because you reported it.

**A (the Windows dev-server break) — fixed in #587**, credited to you. I reproduced both of your findings on macOS using `bun run --shell=bun`, which selects the same shell Windows gets:

```
$ echo port=${PORT:-3002}
port=${PORT:-3002}          # unexpanded even with PORT=9999 set
$ set -a && echo set-worked
bun: command not found: set
```

Exactly your diagnosis, including the root-script `set` issue you flagged as an aside. That second one is the more insidious of the two — `turbo` still runs, so on Windows the root `.env` has silently never loaded and nothing said so. Both scripts now go through `dotenv-cli` with a committed `.env.defaults` loaded last, which keeps the `PORT` override working (verified 3002 by default, 4321 with an override, under both shells) instead of hardcoding it away. Also fixed the `.env.example` 3000-vs-3002 mismatch you noticed.

One correction to my earlier review while I'm here: I claimed Bun doesn't expand `${PORT:-3002}` at all. That was wrong — I'd tested Bun's `$` *template tag*, which escapes interpolated values by design, rather than `bun run` script execution. Under `bun run` with the system shell it expands fine, which is why this only ever broke on Windows. Your original diagnosis was more precise than mine.

**B (the MCP `editorUrl`) was my error** — no bug there. `apps/editor/next.config.ts` already redirects `/editor/:id` → `/scene/:id`, added for exactly this reason, and `/editor/<projectId>` is correct for the hosted app. Details in my comment above.

**C (the autosave data loss) — fixed in #578**, at the hook layer in `packages/editor` rather than the route, so it covers the hosted app and every consumer of the package. The root cause was the stale-baseline problem I described: `createStoredNodeCountTracker` now refreshes explicitly when a graph arrives from storage, so the guard is armed from the first save of the session.

Closing since there's nothing left in this branch to merge. Thanks for the report — a bug that makes the repo un-runnable on a whole platform is high-value, and the console output plus the `set` observation is what made it quick to confirm. Please do send more.

Issue #552, opened 2026-07-27
Discord invite link is ded


krazyjakee:
### What happened?

I click discord link in readme. Is ded :'(

### Steps to reproduce

Click discord link in readme.

### Expected behavior

Discord link is not ded

### Browser & OS

Vibe coded OS & Browser

Ekagrashharma:
where is the link exactly ?

krazyjakee:
https://github.com/pascalorg/editor/blob/42ac4be1ce5f3fee74806aa093267b6fee77d47d/README.md?plain=1#L8

Aymericr:
It was indeed ded. Fixed in #575 — the README now points at `discord.gg/XRKsDcpqgS`, which I verified resolves to the Pascal server and is set to never expire (the old one was a temporary invite that lapsed).

Thanks for including the permalink to the exact line, @krazyjakee — and thanks @Ekagrashharma for chasing it down.

Aymericr:
Landed on `main` in b4abdd8e (#575). The README badge now points at `discord.gg/XRKsDcpqgS`, a non-expiring invite — the previous one was a temporary invite that lapsed, which is exactly the failure mode you hit.

Thanks @krazyjakee for the permalink to the exact line, and @Ekagrashharma for confirming.

PR #575, opened 2026-08-04
fix(mcp): declare returned fields in tool output schemas; add security policy and code of conduct

### LINKED ISSUE #566: `get_level_summary` violates its own outputSchema (`floorIndex`) — fails on every MCP host that calls `tools/list`
`get_level_summary` returns a `floorIndex` field that is not declared in `getLevelSummaryOutput`. Since the MCP SDK generates the JSON Schema with `additionalProperties: false`, any client that has fetched the tool list validates the structured content and rejects the response:

```
MCP error -32602: Structured content does not match the tool's output schema: data must NOT have additional properties
```

This affects every real MCP host (Claude Desktop, Claude Code, Codex, Cursor), because they all call `tools/list` before `tools/call`. It reproduces on an empty default scene — no particular geometry needed.

**Why the test suite misses it:** the SDK client only validates `structuredContent` against a tool's output schema once it has cached that schema from `tools/list`. `packages/mcp/src/tools/scene-query.test.ts` calls `client.callTool()` without a preceding `client.listTools()`, so validation never runs. Adding `await client.listTools()` to the test setup surfaces it immediately.

**Reproduction**

```ts
import { Client } from '@modelcontextprotocol/sdk/client/index.js'
import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js'
import { createPascalMcpServer, SceneBridge } from '@pascal-app/mcp'

const bridge = new SceneBridge()
bridge.loadDefault()
const server = createPascalMcpServer({ bridge })
const [s, c] = InMemoryTransport.createLinkedPair()
const client = new Client({ name: 'repro', version: '0.1.0' })
await Promise.all([server.connect(s), client.connect(c)])

await client.listTools()                                    // <-- what every real host does
await client.callTool({ name: 'get_level_summary', arguments: {} })  // throws -32602
```

Actual payload vs. declared fields:

```
declared: levelId, levelName, role, metadataRole, isOccupiedStory,
          isSupportLevel, referenceLevelId, counts, walls, zones, items, slabs, ceilings
returned: levelId, floorIndex, role, metadataRole, isOccupiedStory,
          isSupportLevel, referenceLevelId, counts, walls, zones, items, slabs, ceilings
extra:    floorIndex
```

**Fix**

`floorIndex` is set in `levelSummary()` at `packages/mcp/src/tools/scene-query.ts:480` and looks intentional, so the schema is what's stale — add to `getLevelSummaryOutput` (`packages/mcp/src/tools/scene-query.ts:38`):

```ts
floorIndex: z.number(),
```

**Scope**

I swept all 37 registered tools with a client that calls `tools/list` first; `get_level_summary` is the only one that violates its output schema. The other 36 are clean.

Checked against `main` @ 08e2279, bun 1.3.11, macOS arm64.

Aymericr:
Two independent commits, both maintainer-side housekeeping.

## 1. MCP tools returned fields their own schemas did not declare

Fixes #566.

`get_level_summary` returns `floorIndex` and `create_from_template` returns `saveSkipped`, and neither was declared in the tool's `outputSchema`. The MCP SDK generates the JSON Schema with `additionalProperties: false`, so once a client has cached the schema from `tools/list` it rejects the response:

```
MCP error -32602: Structured content does not match the tool's output schema: data must NOT have additional properties
```

Every real host — Claude Desktop, Claude Code, Codex, Cursor — calls `tools/list` before `tools/call`, so both tools failed in production on an empty default scene while the whole suite stayed green.

**Why the suite missed it.** The SDK client only validates `structuredContent` against a tool's output schema once it has cached that schema from `tools/list`. No test called `listTools()` first, so validation never ran. @Till-JS diagnosed this precisely in #566 — thank you.

`packages/mcp/src/tools/output-schema-contract.test.ts` connects a client that lists before calling and exercises the read-only tools, so a payload field added without a matching schema entry fails CI from now on. Reverting either schema change makes it fail with the exact -32602 above.

#566 reported `get_level_summary` as the only offender out of 37 tools. That was right for the tools its sweep reached; `create_from_template` only surfaces on the `save: true` + no-store path, which needs a store-less bridge to hit.

## 2. Community health files

- The README Discord badge pointed at an expired invite (#552). Now uses the same non-expiring invite the hosted app already advertises.
- No `SECURITY.md` existed, so there was no stated private channel for a vulnerability report — the only visible options were a public issue or discussion. Adds GitHub private vulnerability reporting plus security@pascal.app, with scope.
- Adds Contributor Covenant 2.1, and a Contributing section in the README linking both files and the plugin guidance `CONTRIBUTING.md` already gives.

## Verification

- `bun test packages/mcp/src` — 299 pass, 0 fail (was 297 pass)
- `bun run check` — 1582 files, clean
- `bun run check-types` — 9/9 tasks pass

---

> [!NOTE]
> **Low Risk**
> Changes are schema declarations, tests, and documentation; MCP behavior is unchanged aside from fixing client-side schema validation failures.
> 
> **Overview**
> **MCP tool output schemas** now declare fields that handlers already return, fixing `-32602` validation failures in hosts that call `tools/list` before `tools/call`. `get_level_summary` adds **`floorIndex`**; `create_from_template` adds optional **`saveSkipped`**.
> 
> A new **`output-schema-contract.test.ts`** runs an MCP client that lists tools first and exercises read-only tools (plus the no-store `save: true` path for templates), so undeclared `structuredContent` fields fail CI.
> 
> **Repo hygiene:** **`next-env.d.ts`** is gitignored and the tracked copy under `apps/ifc-converter` is removed to stop Next dev/typegen churn. **README** updates the Discord invite and adds a **Contributing** section. New **`SECURITY.md`** (private reporting, scope) and **`CODE_OF_CONDUCT.md`** (Contributor Covenant 2.1).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f724952ed361e6ddcd104e46ed260204aa73a24b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Aymericr:
Good catch from Bugbot on the `apps/ifc-converter/next-env.d.ts` churn — that was a `next typegen` side effect from running `check-types` locally.

Rather than just reverting the line, I fixed the cause: the file was tracked in `apps/ifc-converter` but ignored in `apps/editor`, which is why it churns for anyone who runs the typecheck. Next rewrites that import path depending on whether `next dev` or `next typegen`/`next build` touched it last. `f724952` untracks it and moves the ignore rule to the root `.gitignore` so it covers every Next app instead of being duplicated per app. Verified Next regenerates it from scratch and `check-types` still passes 9/9.

Issue #588, opened 2026-08-04
Draw calls scale with wall count: batch a level's static walls per material


Aymericr:
Split out of #492 as cause 2. #556 fixed cause 1 (level-wide recompute per frame); this is the remaining half and a design change rather than a regression.

## Measurement

All numbers from @toycenterboss-bot's instrumented profiling in #492. macOS, Chrome, WebGPU renderer (three r185), `frameloop="never"` — so **50 FPS is a ceiling**, and a 50.x reading means "hit the ceiling", not "hit the hardware limit".

| Scene | Nodes | Contents | Draw calls | Orbit FPS |
|---|---:|---|---:|---:|
| empty | 0 | grid and helpers only | 24 | 50.3 |
| zones | 142 | flat polygons | 130 | 50.2 |
| outer walls | 88 | perimeter only | 534 | 50.0 |
| real plan | 310 | walls and rooms | 1362 | 50.2 |
| full import | 1089 | outer + inner walls | 6520 | 42.7 |

Cost tracks **visible meshes, not node count** — a 142-node zone scene costs 130 calls, an 88-node wall scene costs 534. Every scene under roughly 1400 calls holds the ceiling; the 1089-wall import falls off it.

A scene-graph walk on that scene found 8 meshes with one material group, 1081 with three, and 24 with six: **3395 draw units, doubled by the shadow pass** → 6520 calls.

## Direction

Merging a level's static walls per material collapses those 3243 units to three, which took the same scene to **25 draw calls** in the reporter's experiment.

The constraint to design around: the merge has to invalidate cleanly on edit, and walls are exactly the geometry users mutate most. So the win only holds if the merged batch

- excludes whatever is currently being dragged, and
- rebuilds off the dirty-tracking #556 introduced, rather than a second parallel mechanism.

That ordering is why this is a follow-up to #556 rather than part of it.

Also worth checking while in here: the shadow pass doubling every unit means anything that reduces shadow-casting mesh count pays twice.

## Fixture

A 1089-wall scene JSON reproducing the last row has been offered in #492 and should be added as a profiling fixture so this is measurable in CI rather than by hand.

Credit to @toycenterboss-bot for the profiling and the decomposition — separating "work repeated per frame with unchanged inputs" from "draw calls scale with mesh count" is what made both halves actionable.

Aymericr:
Implemented and merged in #608.

[...]

PR #596, opened 2026-08-05
fix(core): stabilize collinear wall junction ordering

### LINKED ISSUE #581: Wall junction miter order is non-deterministic for exactly-collinear walls
### What happened?

Pre-existing bug, found while reviewing #556 (`perf(core,viewer): stop rebuilding wall geometry every frame`) and promised as a follow-up there. **This is not a regression from that PR** — the caching work in #556 made junction *ordering* load-bearing for cache identity, which is what drew attention to it, but the tie-break gap predates it.

`buildWallJunctions` in `packages/core/src/systems/wall/wall-mitering.ts:279` orders the walls meeting at a junction by their outgoing angle and then miters each adjacent pair:

```ts
// Sort by outgoing angle
processedWalls.sort((a, b) => a.angle - b.angle)
```

The comparator has **no tie-break**. When two walls leave a junction at the *same* angle — `Math.atan2(v.y, v.x)` returning bit-identical values — their relative order is whatever `Array.prototype.sort` produces for the input order. `sort` is required to be stable in modern JS, so the order follows the order walls were pushed into `processedWalls`, which follows scene iteration order. That is not a property we control or should rely on: it changes when a wall is deleted and re-added, when nodes are reordered, when a scene is reloaded from storage, or when a remote peer's mutation lands in a different sequence.

Because the loop miters `processedWalls[i]` against `processedWalls[(i + 1) % n]`, a different order means a different *pairing*, which means different left/right intersection points for those walls — so geometry can differ between two sessions of the identical scene.

### Steps to reproduce

Exactly-collinear same-angle walls at one junction are constructible in a few ways:

1. Two walls of different thickness continuing along one straight line (this shape is already exercised for slab polygons — see `slab-polygon.test.ts:722`, *"two collinear walls of different thickness along one edge"*).
2. A duplicated wall left stacked on top of another after a copy/paste or an interrupted drag.
3. A wall drawn back over an existing one so both share an endpoint and a direction.

Then: save and reload, or delete and undo one of the two, and compare the mitered corner.

### Expected behavior

Junction geometry should be a pure function of the walls at that junction — same walls in, same geometry out, independent of scene iteration order.

### Suggested fix

Give the comparator a deterministic tie-break on a stable intrinsic property, so equal angles resolve the same way every time:

```ts
processedWalls.sort((a, b) => a.angle - b.angle || (a.id < b.id ? -1 : a.id > b.id ? 1 : 0))
```

Node id is stable across reload and across peers, which is what's needed. Thickness would be a tempting secondary key but isn't unique.

Two things worth checking as part of the fix rather than assuming:

- **Is `angle` ever *nearly* equal rather than exactly equal?** A tie-break on strict equality doesn't help two walls that differ by 1e-16 — they'd sort consistently by angle but the pairing is still arbitrary in effect. If near-collinear is the real-world case, the tie-break needs an epsilon, which is a different (and more delicate) change. `wall-mitering.ts:39-46` already documents the near-collinear regime as pathological for miter length, so there's prior context to read.
- **Does the miter cache in `packages/viewer/src/systems/wall/level-miter-cache.ts` key on anything order-dependent?** If it does, non-determinism here can also mean a cache that's valid in one session and stale in the next.

### Additional context

Low user-visible impact — the visual difference at a collinear junction is small, and the shape is uncommon in normal drawing. Filing it because it's a *determinism* bug rather than a rendering bug: order-dependent geometry is the kind of thing that surfaces later as a phantom diff in collaborative editing or an unexplained cache miss, and it's much cheaper to fix now with a one-line comparator than to debug from that end.

Good first issue for someone who wants to work in core geometry with a test to write: a unit test that builds the same junction twice with reversed insertion order and asserts identical output would fail today.

tomatotomata:
## Summary

- make exactly-collinear wall junction ordering deterministic by using wall ID as the stable tie-breaker after outgoing angle
- add a regression test that compares the same wall set in forward and reversed input order

Closes #581

## Validation

- `npx --yes bun@1.3.0 test packages/core/src/systems/wall/wall-mitering.test.ts` passed: 12 tests, 0 failures, 145 expectations
- `npx --yes bun@1.3.0 x biome lint packages/core/src/systems/wall/wall-mitering.ts packages/core/src/systems/wall/wall-mitering.test.ts` passed
- `git diff --check` passed

The near-equal-angle epsilon behavior mentioned in the issue is intentionally out of scope.

---

> [!NOTE]
> **Low Risk**
> Small, localized change to junction sort order in wall rendering geometry; behavior only affects equal-angle cases and is covered by a new test.
> 
> **Overview**
> **Wall miter junctions** at shared corners now sort connected walls by outgoing angle and, when angles match (collinear overlaps), by **wall ID** so adjacent-wall pairing and intersection geometry no longer depend on how walls are ordered in the input array.
> 
> A regression test asserts that `calculateLevelMiters` yields identical `junctionData` for the same wall set in forward vs reversed order at a multi-wall junction.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02aae8daf19a417adb3b1ce9ccd22a26dddf0080. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/systems/wall/wall-mitering.test.ts
+++ b/packages/core/src/systems/wall/wall-mitering.test.ts
@@ -79,6 +79,21 @@ describe('wall miter boundary sides', () => {
   })
 })
 
+describe('deterministic junction ordering', () => {
+  test('same wall set produces the same miter data regardless of input order', () => {
+    const thickWall = (id: string, thickness: number, end: [number, number]) =>
+      ({ ...wall(id, [0, 0], end), thickness }) as WallNode
+    const wide = thickWall('wide', 0.6, [4, 0])
+    const narrow = thickWall('narrow', 0.2, [4, 0])
+    const branch = thickWall('branch', 0.4, [0, 4])
+
+    const forward = calculateLevelMiters([wide, narrow, branch]).junctionData.get('0,0')
+    const reversed = calculateLevelMiters([narrow, wide, branch]).junctionData.get('0,0')
+
+    expect(forward).toEqual(reversed)
+  })
+})
+
 describe('junction grid prefilter', () => {
   function thickWall(
     id: string,
--- a/packages/core/src/systems/wall/wall-mitering.ts
+++ b/packages/core/src/systems/wall/wall-mitering.ts
@@ -357,8 +357,13 @@ function calculateJunctionIntersections(
     }
   }
 
-  // Sort by outgoing angle
-  processedWalls.sort((a, b) => a.angle - b.angle)
+  // Sort by outgoing angle, then by wall ID so equal-angle walls produce the
+  // same pairing regardless of scene iteration order.
+  processedWalls.sort((a, b) => {
+    const angleOrder = a.angle - b.angle
+    if (angleOrder !== 0) return angleOrder
+    return a.wallId < b.wallId ? -1 : a.wallId > b.wallId ? 1 : 0
+  })
 
   const wallIntersections = new Map<string, { left?: Point2D; right?: Point2D }>()
   const n = processedWalls.length


Aymericr:
Approved and merging — thank you @tomatotomata, and welcome. This is the right fix and the right size for it.

Verified rather than assumed:

- **The test genuinely fails on `main`**, and not on map insertion order — I reverted `wall-mitering.ts` to `main` while keeping your test, and it fails with real geometry differences (`wide`'s left corner at `y: 0.3` vs `0.1`, and the mirrored change on `narrow`'s right). That's the pairing flip the issue described, so this test would catch a regression rather than just pass alongside the fix.
- **`wallId` is the correct tie-break key**, and it's what's on `ProcessedWall` — no plumbing needed. Stable across reload and across peers, unlike thickness, which isn't unique.
- **Passthrough walls push two entries with the *same* `wallId`** (`v1` and `v2`), so the comparator can still return `0` for a pair. Not a problem: the two entries are 180° apart, so their angles differ and the tie-break is never reached for them. Worth knowing it's true by construction rather than by luck.
- **`processedWalls.sort` is the only order-sensitive comparator left in the file.** The other one at line 214 keys on an explicit `wallOrder` map, which is already deterministic.
- **The miter cache is safe.** I checked `level-miter-cache.ts` because the issue flagged it: `sameMiterInputs` compares positionally, so a reordered array is a cache *miss*, never a stale hit. With this fix the recompute now returns identical data, which is exactly the property the cache wanted.

Full gate green locally: `check` 1601 files clean, `check-types` 9/9, `test` 12/12 tasks (945 nodes tests + 12 mitering), `build` 7/7.

Agreed on scoping out the near-equal-angle epsilon, and on reflection it belongs out of scope for a stronger reason than "different change": **it isn't a determinism gap at all.** I raised it in the issue as an open question, and working through your fix answers it. `angle` is a pure function of the wall's own geometry, so two walls differing by 1e-16 sort *consistently* every time — the order is stable, it's just not the order a human would call "obvious". What I was actually describing there is miter *quality* at shallow angles, which `MITER_LIMIT` already handles by falling back to a square joint. So `Closes #581` is right and I'm letting it close; no follow-up needed. An epsilon tie-break would also make the comparator non-transitive, which is undefined behavior for `Array.prototype.sort` — so it'd be a bucketing change, not a comparator tweak, and there's no reason to make it.

One note on your validation commands: `npx --yes bun@1.3.0 test <file>` works, but you can run the whole repo gate with `bun run check && bun run check-types && bun run test && bun run build` — that's what CI runs, and it catches cross-package type breakage that a single-file test run won't. Not needed here; your change was clean through all four.

Aymericr:
Merged as `163ef0be`, and #581 closed with it.

One process note for next time: this was open as a **draft**, which is why CI didn't run until I approved the workflow (first-time contributor from a fork needs that regardless) and why the merge initially bounced. The body read as a finished submission — full validation output and an explicit scope note — so I marked it ready myself rather than bouncing it back to you. If a PR is ready for review, opening it undrafted gets it looked at sooner.

Thanks again. If you want another in the same area, #588 (batching a level's static walls per material) is labelled `help wanted` and is the natural next step in the same wall-geometry path you just worked in.

PR #600, opened 2026-08-05
core: shared pure plan-footprint math for spatial-grid and MCP

ActArtech:
## What does this PR do?

Extracts rotation-aware plan (XZ) footprint math into a pure, Node-safe core module so spatial-grid and alignment anchors share one formula (follow-up seam for MCP layout clearance after #569).

- Add `packages/core/src/lib/plan-footprint.ts` with **`planFootprintCorners`** and **`planFootprintAABB`** only (plus `PlanAabb` / `PlanVec2` types)
- Spatial-grid manager and alignment-anchors **delegate** to it (no third parallel AABB path)
- Export `@pascal-app/core/plan-footprint` (pure) and re-export the same two helpers from `@pascal-app/core/spatial-grid`
- Gap call-site meanings (`gap ≈ 0.08` packing vs `gap = 0` hard verify) documented in the module header for the MCP consolidation follow-up
- Unit tests for unrotated / 90° / 45° AABB and corner↔AABB parity

**Narrow response to review:** dropped unused `aabbsOverlapPlan`, `planFootprintAABBForItem`, and `planFootprintAABBFromCorners` (no dead / speculative exports).

**Out of scope (follow-up):** rewiring MCP `door-clearance` / `layout-clearance` / `check-collisions` onto these helpers and deleting MCP's duplicate overlap path.

## How to test

1. `bun test packages/core/src/lib/plan-footprint.test.ts` — all pass
2. `bun test packages/core/src/services/alignment-anchors.test.ts` — still green
3. `bun run check packages/core/src/lib/plan-footprint.ts packages/core/src/lib/plan-footprint.test.ts packages/core/src/hooks/spatial-grid/spatial-grid-manager.ts packages/core/src/services/alignment-anchors.ts` — clean
4. Optional: import `planFootprintAABB` from `@pascal-app/core/plan-footprint` and from `@pascal-app/core/spatial-grid` and confirm the same box

N/A — non-visual change (pure geometry helpers + tests).

## Checklist

Note: this change is pure math with no UI surface; verification was `bun test` on plan-footprint + alignment-anchors, `bun check` on touched files, and green GitHub CI (`quality` / `mcp-ci` / `cli-smoke`).


Aymericr (changes requested):
Thanks for picking up the note I left at the end of #569 — this is the right direction and I checked the math rather than assuming: `planFootprintAABB`'s abs-cos/abs-sin extents are provably the same box as the rotated-corner loop it replaces in `alignment-anchors.ts`, `planFootprintCorners` preserves spatial-grid's rotation convention exactly (different corner order, nothing depends on order), the `attachTo` null-list matches the enum in `item.ts`, and it stays clear of the R3F guard in `architecture.test.ts`. The `alignment-anchors.ts` comment admitting it "mirrors the rotated-corner math the spatial-grid manager uses" was a real duplication and −58 lines is a real removal.

But as it stands the PR moves the number in the wrong direction, and that's what I want to fix before it lands.

**The ask on #569 was that MCP stop being a third copy.** With the rewire declared out of scope, `aabbsOverlap` in `door-clearance.ts` survives — and the new `aabbsOverlapPlan` is a line-for-line reimplementation of it. Post-merge there are three copies of expand-then-intersect instead of two, and MCP still uses its own. The extraction is the easy half; the consolidation was the point.

**Three of the five exports have no callers.** `aabbsOverlapPlan` and `planFootprintAABBForItem` exist for the deferred MCP work; `planFootprintAABBFromCorners` is called only by tests, which then assert it's identical to `planFootprintAABB` — so it's an identity helper proving itself. On #569 I asked you to delete `aabbFromPlan` and `nodeItemAabb` for exactly this, and `AGENTS.md` names dead code and speculative abstractions specifically.

**Two paths, either fine by me:**

- **Finish it.** Keep all five exports and rewire `door-clearance.ts` / `layout-clearance.ts` / `check-collisions.ts` to import from `@pascal-app/core/plan-footprint`, deleting MCP's `aabbsOverlap` and `PlanAabb`. Heads-up on the trap that probably made you defer: `@pascal-app/core` is only a **devDependency** of `packages/mcp` while `door-clearance.ts` already runtime-imports `getScaledDimensions` from `@pascal-app/core/schema` — that's a latent packaging bug, and moving more runtime code across that edge means promoting it to a real dependency. Please also keep the `gap: 0` versus `0.08` distinction from #569 intact per call site; that's the bug I'd most want a test to hold.
- **Or narrow it.** Ship only `planFootprintCorners` and `planFootprintAABB`, drop the other three exports and their tests, and leave the subpath export in as the seam for the follow-up.

On the wiki page: I'd rather it didn't say "Aymericr noted…" — the other `wiki/architecture/` pages state the rule, not who asked for it, and 43 lines of page restating the header of a 133-line math module is more provenance than durable rule. Fold the gap-semantics table into the module header and skip the page, or cut it to the invariant plus the call-site table.

Useful side-finding worth carrying into the rewire: your `aabbsOverlapPlan` comment ("expand each box by `gap`") is right and MCP's existing one ("expand each box by gap/2") describes something the code doesn't do. Fix that comment while you're in there.

ActArtech:
Took the **Narrow** path from the review:

- Public surface is now only `planFootprintCorners` + `planFootprintAABB` (plus `PlanAabb` / `PlanVec2` types).
- Dropped `aabbsOverlapPlan`, `planFootprintAABBForItem`, and `planFootprintAABBFromCorners` with their tests.
- Removed `wiki/architecture/plan-footprint.md` (no named attribution). Gap call-site meanings (`0` vs `~0.08`) live in the `plan-footprint` module header as the seam for the MCP consolidation follow-up.
- `@pascal-app/core/plan-footprint` subpath export kept.

Follow-up (not in this PR): rewire MCP `door-clearance` / `layout-clearance` / `check-collisions` onto these helpers and delete MCP's duplicate overlap path.

Aymericr (approved):
The narrowed version now exposes only the two shared footprint primitives and removes the speculative overlap helpers and wiki attribution. Against current main I verified 141 focused footprint/alignment/spatial-grid tests, Biome, and the core build.

PR #321, opened 2026-05-20
fix: respect imperial units in wall panel

MateoSaettone:
## Summary
- add shared linear measurement helpers for meter/imperial formatting and conversion
- reuse the shared formatter across wall, fence, site edge, item placement, and floorplan measurement labels
- refactor the existing imperial-aware MetricControl to use the shared conversion helpers instead of its own feet multiplier
- display/edit wall metadata dimensions in ft when the viewer unit is imperial while keeping scene values in meters
- clamp final stored meter values after display-unit rounding so imperial controls cannot escape physical bounds
- harden shared measurement formatting for non-finite and signed values

FixedImage

After the fix, imperial mode keeps the selected wall metadata panel in `ft` while scene values continue to persist internally in meters.

Closes #322

## Test Plan
- `bun test packages/editor/src/lib/measurements.test.ts`
- `bunx biome check packages/editor/src/lib/measurements.ts packages/editor/src/lib/measurements.test.ts packages/editor/src/index.tsx packages/editor/src/components/editor/wall-measurement-label.tsx packages/editor/src/components/editor/site-edge-labels.tsx packages/editor/src/components/editor/floorplan-panel.tsx packages/editor/src/components/tools/item/use-placement-coordinator.tsx packages/editor/src/components/ui/controls/metric-control.tsx packages/nodes/src/wall/tool.tsx packages/nodes/src/wall/panel.tsx packages/nodes/src/fence/tool.tsx`
- `bun run --cwd packages/editor check-types`
- `bun run check-types`
- `bun run --cwd packages/nodes build`

## Notes
- Editable numeric controls continue to use decimal feet (for example, `10.50 ft`); passive measurement labels continue to use feet/inches (for example, `10'6"`).
- Follow-up issue #374 tracks remaining non-wall measurement panels that still need imperial-aware editable controls.

PR #602, opened 2026-08-05
feat: add exact wall-length input during drafting

### LINKED ISSUE #308: unit toggle only affects the display of dimensions, not the input

aryansk:
## Summary

- Add an exact-length input during wall drafting in both the 2D and 3D flows.
- Parse the entered measurement with the existing unit-aware wall measurement behavior.
- Keep the live wall measurement and unit toggle behavior aligned with the entered value.

## Validation

- 596 editor tests
- Editor typecheck
- Nodes, core, and viewer builds
- Biome
- git diff --check

Closes #308

### DIFF
--- a/packages/editor/src/components/editor/floorplan-panel.tsx
+++ b/packages/editor/src/components/editor/floorplan-panel.tsx
@@ -182,8 +182,10 @@ import {
 } from '../tools/stair/stair-defaults'
 import {
   chainEndJoinsExistingWall,
+  constrainWallDraftLength,
   createWallOnCurrentLevel,
   isSegmentLongEnough,
+  parseWallDraftLength,
   resolveTerrainWallConstructionOptions,
   snapWallDraftPoint,
   snapWallDraftPointDetailed,
@@ -2374,6 +2376,9 @@ type DraftWallMeasurement = {
 }
 
 function FloorplanDraftWallMeasurement({
+  lengthInput,
+  lengthInputError,
+  onLengthInputChange,
   measurement,
   measurementStroke,
   labelBackground,
@@ -2381,6 +2386,9 @@ function FloorplanDraftWallMeasurement({
   sceneRotationDeg,
   unitsPerPixel,
 }: {
+  lengthInput: string
+  lengthInputError: boolean
+  onLengthInputChange: (value: string) => void
   measurement: DraftWallMeasurement
   measurementStroke: string
   labelBackground: string
@@ -2415,7 +2423,8 @@ function FloorplanDraftWallMeasurement({
   const cx = measurement.midpoint[0] + perpX * offset
   const cy = measurement.midpoint[1] + perpY * offset
 
-  const lengthTextWidth = measurement.lengthLabel.length * upx * 6.2
+  const lengthTextWidth =
+    Math.max(measurement.lengthLabel.length, lengthInput.length, 8) * upx * 6.2
   const lengthPlateW = lengthTextWidth + padX * 2
   const lengthPlateH = fontSize + padY * 2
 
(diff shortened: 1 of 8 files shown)


Aymericr:
Sorry for the silence on this — it's a draft so it fell off my radar, and it shouldn't have. This is the feature #308 has been waiting for and the foundations are right: heading from the pointer, distance from the input is the correct split, you reused `parseMeasurement` with `bareUnit` following the unit toggle instead of re-deriving the conversion, and you landed it in both the 2D `floorplan-panel` and the 3D `wall/tool.tsx` in one PR, which most attempts at this miss. `constrainWallDraftLength([0,0],[3,4],2) → [1.2,1.6]` with a test is exactly the shape I wanted.

Four things on the interaction before it can land — all above the geometry, none of it a rewrite:

1. **Don't mount the input unconditionally.** Right now a focusable text field with `pointerEvents: 'auto'` and `stopImmediatePropagation` on pointerdown sits at the segment midpoint — on top of the wall being drawn. A click meant to place the endpoint can land in the field instead. Keep `DraftMeasurementLabel` as the default and have a digit keypress swap it for the input; that's also the pattern people arrive from other CAD tools expecting, since you type rather than aim.
2. **Give it Enter and Escape.** `onKeyDown={(e) => e.stopPropagation()}` is unconditional, so with focus in the field there's no way to commit the length and no way to cancel the tool. Enter should place the segment; Escape should clear the field on first press and fall through to the tool on second.
3. **Decide what happens to snapping, and show it.** Forcing `useWallSnapIndicator` to `null` whenever a length is set means corner and T-junction snapping silently stop, with nothing on screen saying so. I'd rather the length constrain the *distance* and leave the angle snap live, so typing `3m` along a snapped axis still snaps the axis. If you'd rather keep the current rule, the HUD needs to indicate that the endpoint is length-locked.
4. `DraftMeasurementLabel` is now unused for walls but still drives fence and roof, so the wall HUD diverges from its siblings. Either keep walls on the label with the input as an overlay, or note the divergence deliberately.

One practical warning: `floorplan-panel.tsx` has moved a lot since you branched (2D viewer modes in #672–#675, hidden-wall pointer fixes in #683/#689/#697), so expect a real rebase around your `handleWallPlacementPoint` and pointer-move hunks rather than a clean one. Happy to review as soon as it's out of draft — and thanks for reading #308 and #321 before starting, it shows.

Aymericr:
Thanks for the geometry groundwork here. I’m closing this draft because the requested interaction work has remained outstanding since August 28 and the wall-drafting paths on main have moved substantially. Issue #308 stays open: exact-length entry is still wanted. A fresh PR should start from current main and use the interaction already agreed in review: keep the normal measurement label until a digit starts editing, Enter commits, Escape first clears the input and then falls through to cancel drafting, angle/junction snapping remains explicit, and the behavior is shared across the 2D and 3D drafting surfaces.

PR #613, opened 2026-08-08
test: codify space-detection commit visibility

### LINKED ISSUE #586: Reconciliation-generated nodes have no defined path to collaborators
## Problem

Nodes that space-detection reconciliation generates — auto slabs, auto ceilings, auto zones, `frontSide`/`backSide` surface flips, and the `level.children` growth that comes with them — have no defined way to reach collaborators. Anything that moves reconciliation *after* the commit snapshot silently stops replicating.

Today it works by accident of ordering. `initSpaceDetectionSync` subscribes to the store, so its writes land before zundo's `onSave` snapshots the transaction, and the hosted editor's

```ts
// apps/community — use-project-scene-collaboration.ts:552
const diff = deriveSceneOperationChanges(commit.before, commit.current)
```

picks them up inside `commit.current`. Nothing states this as a contract, and nothing tests it.

## Why it needs writing down

#554 restructured reconciliation to run from `subscribeSceneCommits` instead — a reasonable design in isolation, since it gives reconciliation the before/after pair it needs for delta work. But commit listeners run *after* both snapshots are taken, and reconciliation writes under `pauseSceneHistory`, so they emit no commit of their own. Measured on that branch versus `main`, same four-walls-closing-a-room scenario:

| | commit `current.nodes` | live store | never in any commit |
|---|---|---|---|
| `main` | 8 | 8 | — |
| #554 | 6 | 8 | the auto slab + ceiling |

The author gets a floor and a ceiling; collaborators get a bare wall and a divergent `level.children`. That drift doesn't self-heal.

## What would resolve it

1. **State the contract.** Either "reconciliation must write within the triggering commit's transaction" or "reconciliation must emit its own commit that collaboration can transmit." Then document it in `wiki/architecture/` next to the space-detection notes.
2. **Test it at the boundary.** A core-level test asserting that reconciliation-generated nodes appear in the emitted commit would have caught #554 automatically. The check is cheap: close a room, capture commits, assert every live node id appears in some `current.nodes`.
3. **Decide about `pauseSceneHistory`.** Reconciliation pauses history so its writes don't become a separate undo step, which is correct for undo but is also what suppresses the commit. Whether "one undo step" and "one transmittable commit" can be the same thing is the crux.

Option 2 is worth doing regardless of which way 1 goes — right now the invariant is enforced only by call ordering that any refactor can reverse without a single test turning red.

Context: #554 (closed; the `TopologyIndex` half is expected to return as a focused PR).

mvanhorn:
## What does this PR do? / ## How to test / ## Screenshots / screen recording / ## Checklist (bun dev, bun check, docs, main-branch boxes)

Preserve `initSpaceDetectionSync` as a synchronous scene-store subscriber and add a focused rationale at that integration point explaining that reconciliation must finish within the triggering local transaction before its commit snapshot is emitted. Add an architecture page that defines the two-sided replication contract: local reconciliation output is included in the originating commit, while host patches consume those nodes under read-only mode rather than re-running reconciliation. Link that page from the architecture index so future changes to space detection, history pausing, or commit delivery encounter the invariant during design review.

Fixes #586

---

> [!NOTE]
> **Low Risk**
> Tests and documentation only plus an explanatory comment; no changes to reconciliation scheduling or history pausing logic.
> 
> **Overview**
> Locks in how wall-driven room reconciliation must interact with **scene history** and **`SceneCommit`** snapshots, without changing runtime behavior in `initSpaceDetectionSync`.
> 
> Adds an **integration test** on the real `useScene` store: placing the closing wall must yield **one** local commit whose snapshot includes auto slab/ceiling, reconciled wall `frontSide`/`backSide`, and updated level `children`, and **one undo** that removes the wall and generated surfaces together. Documents in **`space-detection.ts`** that reconciliation must stay in the **synchronous** scene-store subscriber (not `subscribeSceneCommits`), because zundo emits the local commit after subscribers return and history-paused derived writes must land inside that boundary.
> 
> Introduces **`wiki/architecture/space-detection.md`** (linked from the architecture index) describing the two-sided contract: local edits reconcile synchronously into the originating commit; peers apply host patches without re-running detection or minting new IDs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b00703cab7712ce7a87adeaedbadee590bf0faa6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/lib/space-detection.test.ts
+++ b/packages/core/src/lib/space-detection.test.ts
@@ -3,7 +3,11 @@ import { BuildingNode, CeilingNode, LevelNode, SlabNode, WallNode, ZoneNode } fr
 import type { AnyNode, AnyNodeId } from '../schema/types'
 import { resolveCeilingHeight } from '../services/level-height'
 import { getCeilingClampBound } from '../services/storey'
-import { runWithSceneCommitNodeIds, type SceneCommit, subscribeSceneCommits } from '../store/history-control'
+import {
+  runWithSceneCommitNodeIds,
+  type SceneCommit,
+  subscribeSceneCommits,
+} from '../store/history-control'
 import useScene, { clearSceneHistory } from '../store/use-scene'
 import {
   detectSpacesForLevel,
--- a/packages/core/src/lib/space-detection.test.ts
+++ b/packages/core/src/lib/space-detection.test.ts
@@ -3,7 +3,8 @@ import { BuildingNode, CeilingNode, LevelNode, SlabNode, WallNode, ZoneNode } fr
 import type { AnyNode, AnyNodeId } from '../schema/types'
 import { resolveCeilingHeight } from '../services/level-height'
 import { getCeilingClampBound } from '../services/storey'
-import { runWithSceneCommitNodeIds } from '../store/history-control'
+import { runWithSceneCommitNodeIds, type SceneCommit, subscribeSceneCommits } from '../store/history-control'
+import useScene, { clearSceneHistory } from '../store/use-scene'
 import {
   detectSpacesForLevel,
   initSpaceDetectionSync,
@@ -16,6 +17,16 @@ import {
 } from './space-detection'
 import { encodeTerrainField } from './terrain-codec'
 import { applyHeightPatch, createTerrainField, flattenPatch } from './terrain-field'
+
+type RafFn = (callback: (time: number) => void) => number
+;(globalThis as unknown as { requestAnimationFrame?: RafFn }).requestAnimationFrame ??= (
+  callback,
+) => {
+  callback(0)
+  return 0
+}
+;(globalThis as unknown as { cancelAnimationFrame?: (id: number) => void }).cancelAnimationFrame ??=
+  () => {}
 
 const square: Array<[number, number]> = [
   [0, 0],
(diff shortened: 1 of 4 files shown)


Aymericr (changes requested):
This is what #586 was asking for, and the diagnosis is right — the comment at `initSpaceDetectionSync` names the actual mechanism (zundo snapshots after subscribers return), and the test is the guard that would have gone red on #554 instead of us catching the missing slab and ceiling by measurement. The wiki page covering both directions, including the `origin: 'host'` / `readOnly` half, is more than I expected.

Three things before I merge:

1. **Rebase.** `space-detection.ts` has grown about 750 lines since you branched and the anchor moved, and `space-detection.test.ts` now imports `runWithSceneCommitNodeIds` from `../store/history-control`, so your new import from that module conflicts. Mechanical, but please confirm the test still passes after.
2. **Say why the real store.** Every other test in that file uses the minimal store stand-ins. Driving the real `useScene` singleton is the right call here — a stub can't exercise the zundo boundary you're asserting — but it looks like an oversight without a line saying so. Please also restore `useScene` state in the `finally`, not just `clearSceneHistory()`, so the singleton mutation can't leak into a later test.
3. **Retitle to `test:`** — there's no behavior change in the diff, and `fix:` will read wrong in the changelog.

One note so it doesn't surprise you: my follow-up on #586 widened the ask to an equivalence harness (scoped reconciliation must produce the same spaces and surfaces as a full `detectSpacesFromWalls`). That's for the `TopologyIndex` work when it returns, not a condition on this. Your commit-visibility assertion stands on its own and I want it in `main` now.

mvanhorn:
All three done.

**Rebased.** Onto current main (81 commits). Two conflicts, both from the growth you flagged: `space-detection.test.ts` now also imports `runWithSceneCommitNodeIds` from `history-control`, merged with the `subscribeSceneCommits` / `SceneCommit` imports; and in `space-detection.ts` the new `adoptSceneBaseline` / non-local commit block landed where my comment goes, so the comment now sits directly above `sceneStore.subscribe` where it explains that choice. Confirmed after: **62/62 in `space-detection.test.ts`**.

**Said why the real store.** The test now carries a line stating that the real `useScene` singleton is required to exercise zundo's commit snapshot boundary, since a stand-in cannot. It also captures `useScene.getState()` up front and restores it with `useScene.setState(previousSceneState, true)` in the `finally`, alongside `clearSceneHistory()`, so the singleton mutation cannot leak into a later test.

**Retitled to `test:`.** Agreed on the changelog reading — there is no behavior change in the diff.

Noted on the equivalence harness being scoped to the `TopologyIndex` work rather than this PR.

Aymericr:
Verified the rebase — parent is `67ea40ef`, single commit, and both conflict areas came through clean: one `adoptSceneBaseline`, one non-local commit block, and the comment sitting directly above `sceneStore.subscribe` where it now belongs (space-detection.ts:2430). The `finally` restore with `useScene.setState(previousSceneState, true)` is exactly what I asked for, and the one-line justification at the top of the test reads right.

One thing left, and it's yours: the red quality check is formatting only. The import you hand-merged at `packages/core/src/lib/space-detection.test.ts:6` is 109 chars against our `lineWidth: 100`, and biome wants it split. `bun run check --write` (or just break the import) fixes it. It also matters beyond lint: `bun run check` runs before `bun run test` in the quality job, so CI has never actually executed the suite on this head — once the format fix lands and quality goes green, that's the 62/62 confirmation I need.

I retitled the PR to `test:` to match the commit (with a second commit on the branch, the squash title comes from the PR title, and `fix:` would have snuck back into the changelog).

Fix the line and I'll merge.

mvanhorn:
Fixed in b00703c — `bun run check --write` split the history-control import.

Quality is green now, which means the suite actually ran on this head for the first time: 62 pass, 0 fail, 241 expect() calls. That's the 62/62 you were waiting on.

Thanks for the retitle catch on the squash title, and for verifying the rebase line by line.

Aymericr (approved):
The requested rebase, real-store isolation, rationale, and formatting fix are present. I re-ran the focused space-detection suite against current main: 62 passed with 241 assertions; the core build and Biome checks also pass.

PR #615, opened 2026-08-09
fix(core): keep numeric sanitize diagnostics fail-safe

anton-pascal:
## Sentry issues

- `MONOREPO-EDITOR-MK` — `TypeError: Cannot read properties of undefined (reading 'join')`
- `MONOREPO-EDITOR-MJ` — related same-second downstream render failure; this change addresses the shared invalid-update path but does not claim a separate MJ fix.

## Root cause

An invalid numeric node update correctly entered the sanitizer, but its warning formatter assumed every runtime `NumericSanitizeIssue.path` was an array and called `.join()` unconditionally. That made the error reporter throw inside `updateNodesAction`, masking the original schema-validation failure and aborting the remaining update batch. `NumericSanitizeIssue.path` is statically required, and all current `sanitizeNumericValue` call paths start from `[]` and extend with array spreads, so no reproducible in-source producer of an undefined path was found; the Sentry payload demonstrates the runtime invariant was nevertheless violated.

## What changed

- make numeric issue formatting tolerate null/undefined issue lists and missing/non-array paths, with a final non-throwing fallback
- isolate both message construction and the warning sink so diagnostics can never interrupt node mutation
- add regressions for missing paths and for a batched `updateNodes` call continuing after schema-invalid numeric data even when warning output throws

## Verification

- `bun test packages/core/src/store/actions/node-mutation-sanitize.test.ts` — PASS: 11 passed, 0 failed
- `bun biome lint packages/core/src/store/actions/node-actions.ts packages/core/src/store/actions/node-mutation-sanitize.test.ts` — PASS: checked 2 files, no fixes applied
- `git diff --check` — PASS

This is unverified against production traffic and needs review before merge.

---

> [!NOTE]
> **Low Risk**
> Localized to diagnostic logging around existing sanitization; behavior change is fail-safe continuation of mutations with added defensive formatting.
> 
> **Overview**
> Hardens **numeric sanitization warning** formatting so bad diagnostic data cannot abort node create/update batches (fixes Sentry `path.join` on undefined).
> 
> **`numericSanitizeIssuesToMessage`** is exported and now accepts null/undefined issue lists, treats missing or non-array `path` as `<unknown>`, and falls back to `<diagnostic unavailable>` instead of throwing.
> 
> **`warnSanitizedNodeMutation`** wraps both message building and `console.warn` in try/catch so a broken formatter or logging sink never interrupts mutations.
> 
> Adds tests for defensive issue formatting and for **`updateNodes`** applying later updates when `console.warn` throws after an invalid numeric patch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit def7b3157fcb4416847b46fbebe1649cdc1ff103. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/store/actions/node-actions.ts
+++ b/packages/core/src/store/actions/node-actions.ts
@@ -459,14 +459,24 @@ function formatNumericValue(value: number) {
   return String(value)
 }
 
-function numericSanitizeIssuesToMessage(issues: NumericSanitizeIssue[]): string {
-  return issues
-    .map((issue) => {
-      const path = issue.path.map(String).join('.') || '<root>'
-      const to = issue.to === undefined ? '' : ` -> ${formatNumericValue(issue.to)}`
-      return `${path}: ${formatNumericValue(issue.from)} ${issue.action}${to}`
-    })
-    .join('; ')
+export function numericSanitizeIssuesToMessage(
+  issues: NumericSanitizeIssue[] | null | undefined,
+): string {
+  if (!Array.isArray(issues)) return ''
+
+  try {
+    return issues
+      .map((issue) => {
+        const path = Array.isArray(issue?.path)
+          ? issue.path.map(String).join('.') || '<root>'
+          : '<unknown>'
+        const to = issue?.to === undefined ? '' : ` -> ${formatNumericValue(issue.to)}`
+        return `${path}: ${formatNumericValue(issue?.from)} ${issue?.action ?? 'sanitized'}${to}`
+      })
+      .join('; ')
+  } catch {
+    return '<diagnostic unavailable>'
+  }
 }
 
 function warnSanitizedNodeMutation(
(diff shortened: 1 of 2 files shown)


anton-pascal:
Status note: this is still based on `27adf9a0`, which is now **10 commits behind** `main` (`34fbae74`). The bug this fixes is **still present on `main`** — `numericSanitizeIssuesToMessage` still calls `issue.path.map(String).join('.')` with no array guard, so a malformed `NumericSanitizeIssue` still throws inside the error reporter and aborts `updateNodesAction` (MONOREPO-EDITOR-MK). GitHub reports this branch `mergeable: true` / `clean`, so it can still land, but the downstream app-side bump (pascalorg/private-editor#317) has gone stale: private-editor's submodule pointer moved to `f90c7414`, which is 7 commits *ahead* of this branch's head, so that bump would now roll the submodule backwards. I've converted #317 to draft. Cleanest path is to rebase this onto current `main`, land it, then re-point the submodule forward. Not rebasing unprompted — @Aymericr's call.

PR #648, opened 2026-08-14
feat(editor): add undo/redo buttons to viewer toolbar

rkolpakov-sudo:
## What does this PR do?

Adds Undo/Redo buttons to the viewer toolbar (top-left, next to the collapse and view-mode controls). The buttons subscribe to the history store via `subscribeHistoryCommandState` and dispatch through `runUndo`/`runRedo`, which respect the collaborative history delegate and fall back to standalone Zundo when no controller is installed. Buttons disable when nothing can be undone/redone.

Keyboard shortcuts were already wired (Ctrl/Cmd+Z undo, Ctrl/Cmd+Shift+Z redo) and command palette entries already existed; this PR adds the visible toolbar affordance and the CHANGELOG entry.

## How to test

1. Run `bun dev` and open the editor at http://localhost:3002
2. Create or move a node in the 3D view
3. Click the Undo button — the action is reverted; click Redo — it is reapplied
4. With an empty history, both buttons are disabled
5. Press Ctrl/Cmd+Z and Ctrl/Cmd+Shift+Z to confirm the keyboard shortcuts still work alongside the buttons

N/A — visual change; a screen recording will be added if requested.

## Checklist

---

> [!NOTE]
> **Low Risk**
> UI-only addition that reuses existing history APIs; no changes to undo/redo or scene mutation logic.
> 
> **Overview**
> Adds **Undo** and **Redo** controls to the top-left community viewer toolbar (between sidebar collapse and 3D/2D/split view mode).
> 
> A new `HistoryControls` block subscribes to `subscribeHistoryCommandState`, reads `canUndo` / `canRedo` via `getHistoryCommandState`, and invokes `runUndo` / `runRedo` so behavior matches the command palette and keyboard shortcuts (including collaborative history when a delegate is set). Buttons use the same toolbar styling as other controls and are visually disabled when history is empty. **CHANGELOG** documents the feature under Unreleased → Features.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55a63b694287ce10e8b46f95d5ed000bc821ccb2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Features
+
+- Undo/Redo buttons in the viewer toolbar (top-left, alongside the collapse and view-mode controls). The buttons subscribe to the history store via `subscribeHistoryCommandState`, dispatch through `runUndo`/`runRedo` (respecting the collaborative history delegate), and disable when nothing can be undone/redone. Keyboard shortcuts were already wired: Ctrl/Cmd+Z undo, Ctrl/Cmd+Shift+Z redo.
+
 ### Fixes
 
 - Preserve custom scene materials across save, load, clone, fork, and live sync. Materials were dropped at every persistence boundary, so a scene reopened with default surfaces. Collections were dropped on MCP import for the same reason ([#597](https://github.com/pascalorg/editor/pull/597)) by [@ShiroKSH](https://github.com/ShiroKSH)
--- a/apps/editor/components/viewer-toolbar.tsx
+++ b/apps/editor/components/viewer-toolbar.tsx
@@ -10,6 +10,10 @@ import {
   DropdownMenuSubContent,
   DropdownMenuSubTrigger,
   DropdownMenuTrigger,
+  getHistoryCommandState,
+  runRedo,
+  runUndo,
+  subscribeHistoryCommandState,
   useEditor,
   useFloorplanAnnotationVisibility,
   useFloorplanMode,
@@ -37,6 +41,7 @@ import {
   Layers3,
   Magnet,
   PenLine,
+  Redo2,
   Ruler,
   ScanLine,
   SlidersHorizontal,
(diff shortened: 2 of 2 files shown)


Aymericr (changes requested):
Thanks for this, and apologies for the slow response. The feature is one we want, and you found the right seam — routing through `runUndo`/`runRedo` rather than reaching into the history store directly is exactly right, and it means the buttons honor the collaborative delegate for free.

One structural change before it lands. `apps/editor` is our standalone app; the hosted editor has its own toolbar, so as written this ships to one surface and the hosted one still has no undo button. Could you move `HistoryControls` into `packages/editor` — it belongs next to `src/lib/history.ts`, alongside the command-palette entries that already call the same API — and export it from the barrel? Then both toolbars can use one implementation. You'll need to either lift `TOOLBAR_CONTAINER`/`TOOLBAR_BTN` alongside it or take them as props; either is fine by me.

Two smaller things while you're in there:
- Please use `useSyncExternalStore` instead of `useState` + `useEffect`. It's the pattern we use everywhere else for this (`use-plugin-panels.tsx`, `plugins-panel.tsx`, `site-panel/index.tsx`), and it closes the gap where an update between first render and effect commit gets dropped.
- Use real `disabled` rather than `aria-disabled` plus a guard in the handler — the guard works, but the button stays focusable and clickable, and `disabled` is simpler.

For the CHANGELOG, our entries end with the PR link and your attribution, and stay at the user-facing level — no need to name `subscribeHistoryCommandState`. Have a look at the neighbouring entries for the shape.

Last thing: right now it sits between the sidebar collapse and the view-mode switcher, which puts history in the middle of the view controls. I'd rather it were its own group — let me know if you want a suggestion on placement.

Aymericr:
Thanks again for finding the correct history command seam. I’m closing this branch after the requested changes remained unaddressed: the controls still live only in the standalone app, use an effect-backed subscription instead of `useSyncExternalStore`, and emulate disabled state instead of using the native button behavior.

The feature remains welcome. A fresh PR should put one exported `HistoryControls` implementation in `packages/editor` so both standalone and hosted toolbars can use it, then add each host’s placement separately. That narrower shape should be straightforward to review.

PR #726, opened 2026-08-28
fix(editor): store scans locally without host uploads

### LINKED ISSUE #645: Self-hosted editor: scan upload button is a silent no-op (onUploadAsset never wired locally)
## Environment

- `@pascal-app/cli` 0.1.5, self-hosted via `npx @pascal-app/cli start` (Linux, systemd), browser on `http://pascal.localhost:<port>`
- No cloud account involved — purely the local editor + local SQLite storage

## What happens

In the References panel (Reference settings → Scans → "Upload scan or guide image"), picking a scan file (`.glb`) is a **silent no-op** in the self-hosted editor:

- the file chooser opens and accepts the file,
- then nothing: no network request is issued, no console error, no upload progress, no scan node added ("1 scan on this level" stays unchanged).

Verified with devtools/network open: zero requests after file selection.

## Why (from source)

`packages/editor/src/components/ui/sidebar/panels/site-panel/index.tsx` — `handleUpload` ends with:

```ts
clearUpload(levelId)
onUploadAsset?.(projectId, levelId, file, type)
```

`onUploadAsset` is an optional prop and nothing in the local runtime provides it, so the optional call vanishes. (Guide images are fine — they have a local in-tree path with real error handling right above; it's specifically the scan branch that depends on the host-app callback.)

So the button is effectively cloud-only today, but it still renders and opens a chooser in the self-host build — which reads as "broken" rather than "unavailable".

## Expected

Either of:

1. A local default implementation for the self-host runtime (store the file in local asset storage, create the scan node pointing at it) — clearly the nicer fix, since MCP-created scan nodes with a URL already render fine locally, or
2. At minimum: hide/disable the scan upload button (with a hint) when no `onUploadAsset` is wired, instead of silently dropping the file.

## Workaround we use

Serve the GLB from a local static file server and create the scan node via the MCP `apply_patch` tool with that URL — works, but is invisible to normal users of the UI.

Happy to attempt a PR for option 1 if you can hint at the intended local asset-storage path for uploads.

anisayakmitra-in:
## Summary

Fixes the self-hosted scan upload path, which currently accepts a file and then silently does nothing when a host upload callback is absent.

## Details

- Stores local GLB and glTF uploads through the existing IndexedDB asset storage.
- Creates and selects the corresponding ScanNode on the active level.
- Retains the existing host upload flow when a callback is supplied.
- Adds coverage for local scan asset creation.

Fixes #645

## Validation

- bun test packages/editor/src/lib/local-guide-image.test.ts
- bun run --filter @pascal-app/editor check-types
- bun run --filter @pascal-app/editor test (760 pass)
- biome check on the changed implementation files

---

> [!NOTE]
> **Medium Risk**
> Changes cut across npm OIDC publishing, CLI runtime download integrity, hosted MCP registry contracts, and monorepo package layout—high impact if misconfigured, though largely additive CI and distribution work.
> 
> **Overview**
> This PR **ships the 1.0.0 release narrative and tooling** alongside substantial product and distribution changes—not the local scan-upload fix described in the current PR template.
> 
> **Agent workflows and MCP distribution** — Adds versioned **`pascal-3d`** and **`furniture-fit`** skill bundles with marketplace manifests for Claude Code, Codex, Cursor, and Gemini, portable **`mcp.json`**, branding assets, and **`bun run skills:validate`** (wired into CI). A new **MCP Registry** workflow validates **`server.json`** against the live hosted endpoint and API catalog. Internal repo skills gain **`metadata.internal`**; **`open-pr2`** is a separate, explicit PR-writing workflow; **`review-architecture`** expands guidance on animation-driven dirty marking.
> 
> **CLI and npm release pipeline** — **`@pascal-app/cli`** becomes a small npm package with MCP bundled in-package while the **web editor runtime** downloads from a GitHub release asset with **SHA-256 verification**; release uploads that asset after publish. Release workflow drops **`capture-protocol`** / **`capture-viewer`** packages, switches npm publish to **GitHub OIDC trusted publishing** (no **`NPM_TOKEN`**), fixes prerelease graduation semver, and adds **`pascal agent claim`** / **`agent status`** docs in the changelog.
> 
> **Package consolidation** — Standalone **`@pascal-app/capture-protocol`** and **`@pascal-app/capture-viewer`** packages are removed; capture lives under **`@pascal-app/core/capture`** and **`@pascal-app/viewer/capture`**, with README/AGENTS updates.
> 
> **Standalone editor app** — Bootstraps **Environment** (panel + viewer presentation) and **Pool** plugins, Tailwind sources for Environment, and a reworked **Build** tab: roof **type** picker and **`ToolOptionsPanel`**, **Kitchen** modular cabinet entry, MEP group behavior, fitting thumbnails, and tests for roof feature selection.
> 
> Docs (**CHANGELOG**, **README**, **SETUP**), **three@0.186**, and widespread **1.0.0** version bumps round out the release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 409576ba878c7f86fa4865f3cf889957eb1b7342. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Aymericr (changes requested):
Thanks for this — it's exactly the shape I sketched in #645: mirror `local-guide-image.ts`, persist through `saveAsset`, and let the existing node side do the rendering. I verified the whole chain: `asset://` is on the `AssetUrl` allowlist, `ScanRenderer` resolves it through `useAssetUrl`/`loadAssetUrl` so it survives reload, the 200 MB cap applies before your branch, and the hosted app always passes `onUploadAsset` so the host path is untouched (tightening `onUploadAsset?.()` to `onUploadAsset()` after the guard is right). The test is a faithful mirror of the lib's shape too.

One change before merge, and Bugbot is right about it: the success path needs `setShowScans(true)`. The guide branch right above yours calls `setShowGuides(true)` before selecting; `showScans` is a persisted per-project preference (`use-viewer.ts`), and `packages/nodes/src/scan/renderer.tsx` computes `visible = showScans && node.visible` — so with scans toggled off your upload succeeds invisibly, which is the exact failure mode this PR exists to kill. Grab `setShowScans` next to the `setShowGuides` subscription in `site-panel/index.tsx` and call it in the success path.

Two things I noticed and am explicitly not asking you to fix here:

- Deleting an `asset://` scan orphans the File in IndexedDB (`handleDelete` only routes `http(s)` URLs to `onDeleteAsset`). Guide images have the same gap today, so this stays parity — but scans are up to 200 MB, so I'll file a follow-up for local asset cleanup.
- `mock.module('@pascal-app/core', …)` is process-global in bun test; the spread keeps the blast radius to `saveAsset` and nothing else in the package calls it under test, so it's fine as is.

Push the `setShowScans` fix and this is good to go. Appreciate you carrying this from the issue write-up through to the PR — the workaround you documented there has a proper home now. (I've approved the CI run.)

Aymericr (approved):
The local fallback now stores the scan through the shared asset store and makes scans visible before selecting the new node. I verified the focused regression test, core/viewer builds, editor typecheck, and Biome on the updated branch.

Issue #715, opened 2026-08-24
@pascal-app/editor beta.5 pulls node:module into external browser builds through Manifold print export


smokie40:
### What happened?

Hi Pascal team,

we are integrating @pascal-app/editor into an external Next.js application and found a browser packaging issue in beta.5.

The public @pascal-app/editor root pulls the Manifold print-export path into the browser dependency graph even when print export is not used.

As a result, an optimized Next/Webpack build fails when it reaches `node:module` through `manifold-3d`.

The relevant import chain is:

node:module
→ manifold-3d
→ print-shell-compiler-manifold-core
→ print-shell-compiler-manifold.worker
→ print-shell-compiler-manifold-worker
→ export-manager.tsx
→ @pascal-app/editor root
→ external Next.js application

The build error is:

Module build failed: UnhandledSchemeError:
Reading from "node:module" is not handled by plugins.

The editor APIs and runtime functionality we use otherwise work correctly. The failure is specifically at the optimized browser production-build boundary.

### Steps to reproduce

1. Consume the public `@pascal-app/editor` root from an external Next.js application.
2. Use the current beta.5 source.
3. Import the editor through its public package entrypoint.
4. Run an optimized Next/Webpack production build.
5. The browser bundle traverses the print-export dependency path and eventually reaches `node:module` through `manifold-3d`.
6. The build fails with `UnhandledSchemeError`.

We most recently rechecked exact upstream source SHA:

cfe13068fe9e4ef97c1b94ce24f5a40b44d5fe07

The issue is still present there.

### Expected behavior

The public browser editor entrypoint should be consumable by an external browser application without pulling Node-only dependencies into the browser bundle.

Either of these approaches would solve our integration issue:

1. Isolate/lazy-load the print/Manifold implementation so it is not part of the editor browser dependency graph unless print export is actually used.

or

2. Expose a supported browser-safe editor/embed package subpath for host applications that need the editor composition APIs but not print export.

For example, a public entrypoint conceptually similar to:

@pascal-app/editor/embed

containing the browser-safe editor surface would also work for us.

### Browser & OS

N/A — the failure happens during the optimized Next.js/Webpack production build, before browser runtime.

### Additional context

This does not appear to be a general editor/runtime incompatibility.

Our integration checks for the editor API, runtime/client geometry, dimensions, workspace adapter, and ordinary Web tests pass. The blocker is specifically the production browser bundle.

[...]

PR #727, opened 2026-08-28
feat(editor): expose printable STL export

nicemit:
## What does this PR do?

Exposes printable STL export in the settings panel Export section. Users can now download printable STL files (as a zip) via a dedicated button, using the same safe print profile as the existing 3MF export (`printScale: 100`, `printScope: levels`, `printContent: structure`, `printBase: none`).

Follow-up to [Discussion #331](https://github.com/pascalorg/editor/discussions/331) — exposes the existing internal `print-stl` export path in the Settings UI.

## How to test

1. Run `bun dev` and open http://localhost:3002
2. Open or create a project with printable geometry
3. Open the sidebar → **Export** section
4. Confirm both **Export 3D print 3MF** and **Export 3D print STL files** buttons are visible
5. Click **Export 3D print STL files** — a zip should download
6. Click **Export 3D print 3MF** — still works as before
7. Run `bun --cwd packages/editor run test -- print-export-button`

Screenshot_20260828_140351

Screenshot shows the new "Export 3D print STL files" button alongside the existing 3MF export in the Export panel.

## Checklist

---

> [!NOTE]
> **Low Risk**
> UI-only wiring to an existing `print-stl` export path with unchanged preflight and export options; no auth or data-model changes.
> 
> **Overview**
> Adds a second **Export 3D print STL** control next to the existing 3MF button in the settings Export section, so users can download printable STL (zip) from the same UI.
> 
> `preparePrintExport` now takes an explicit `print-3mf` or `print-stl` format instead of always calling `print-3mf`; both paths still use the same fixed safe print options (scale, levels scope, structure content, no base). Export loading state tracks which format is running so one export disables both buttons and sets per-button `aria-busy`. Tests pass the format argument and assert STL uses the same profile as 3MF.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e67e701cce589141329841f0c3b588e453248d87. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/editor/src/components/ui/sidebar/panels/settings-panel/print-export-button.test.ts
+++ b/packages/editor/src/components/ui/sidebar/panels/settings-panel/print-export-button.test.ts
@@ -58,6 +58,33 @@ describe('simple 3D print export', () => {
     expect(prepared).toEqual({ artifact, report })
   })
 
+  test('uses the same safe profile for printable STL files', async () => {
+    const calls: { format?: string; options?: ModelExportOptions }[] = []
+    const stlReport: PrintExportReport = { ...report, format: 'stl' }
+    const artifact = { blob: new Blob(['stl']), filename: 'house.zip', metadata: stlReport }
+    const modelExport: ModelExport = async (format, options) => {
+      calls.push({ format, options })
+      return artifact
+    }
+
+    const prepared = await preparePrintExport(modelExport, false, 'print-stl')
+
+    expect(calls).toEqual([
+      {
+        format: 'print-stl',
+        options: {
+          onlyVisible: false,
+          download: false,
+          printScale: 100,
+          printScope: 'levels',
+          printContent: 'structure',
+          printBase: 'none',
+        },
+      },
+    ])
+    expect(prepared).toEqual({ artifact, report: stlReport })
+  })
+
   test('blocks the download when preflight finds invalid geometry', async () => {
     const blockedReport: PrintExportReport = {
       ...report,
--- a/packages/editor/src/components/ui/sidebar/panels/settings-panel/print-export-button.test.ts
+++ b/packages/editor/src/components/ui/sidebar/panels/settings-panel/print-export-button.test.ts
@@ -40,7 +40,7 @@ describe('simple 3D print export', () => {
       return artifact
     }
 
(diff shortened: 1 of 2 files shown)


Aymericr (changes requested):
Thanks — this is exactly the follow-up I left open in #331, and it's shaped the way I hoped: both actions stay on the fixed safe profile, preflight still gates the download, 3MF stays first as the recommended default, and the new test locks the STL path to the same profile the 3MF test does. I also checked this against #715 (the manifold/webpack module-graph problem): you don't touch the import chain, so it's neutral there — the browser-safe boundary is a separate track on my side, and nothing here makes it harder.

Two small things before I merge:

1. `print-export-button.tsx` — `aria-busy` now lies on the idle button: both buttons share `isExporting`, so exporting 3MF announces the STL button as busy too. Track the in-flight format (`useState<PrintModelExportFormat | null>`) and set `aria-busy` only on the active button. Keep both *disabled* during an export — that part is right, since they share one export pipeline.

2. Labels: "Export 3D print 3MF" vs "Export 3D print STL files" isn't parallel, and the STL artifact is actually a zip of per-level STLs (`print_levels_1-100_<date>.zip`). Make them "Export 3D print 3MF" / "Export 3D print STL" — the plural/zip detail is better carried by the downloaded filename than the button. Worth keeping them clearly distinct from the plain "Export STL" button that already lives in this panel, which "3D print" does.

The `format` default on `preparePrintExport` is only there so the old test call sites don't change — I'd make it a required parameter, but I won't hold the PR on that.

I checked for anything depending on the old "Export 3D print files" label — nothing in tests or docs references it, so the rename is safe. I've approved the workflow run; once the two items above are in I'll merge.

nicemit:
Thanks for the review @Aymericr, addressed both items (plus making `format` required on `preparePrintExport`):

1. `aria-busy` is scoped to the in-flight format; both buttons stay disabled during export
2. Label is now "Export 3D print STL"

Aymericr (approved):
The requested accessibility state and parallel labels are fixed. I verified the focused print-export tests, core/viewer builds, and editor typecheck against current main; both formats retain the same fixed preflight profile.

PR #729, opened 2026-08-28
Carry scene materials through Load Build

alxbouchard:
Split out of #720 as requested in the review — the materials fix on its own, so it doesn't wait on the import-page discussion.

## The bug

`validateBuildJson` drops the top-level `materials` table: every `scene:<id>` slot ref in an imported file points at a material that no longer exists, so custom finishes silently revert to defaults on Load Build. Confirmed against main: `handleConfirmImport` passes only `installedPlugins` to `setScene`, whose `extra.materials` support already exists.

## The fix

- `ParsedBuildJson` gains `materials?: Record<string, SceneMaterial>`; entries are validated one by one (`SceneMaterial.safeParse`), invalid ones skipped with a warning so a bad material never takes the import down.
- **Normalization is a deliberate, documented choice** (per the review): `safeParse().data` injects defaults and drops unknown keys — the opposite of `apiGraphSchema`'s preserve-unknowns stance — because import feeds the live scene store, which only understands schema-shaped materials. The in-line comment states this and why.
- `handleConfirmImport` now takes `ParsedBuildJson` directly (no widened record + cast) and hands `parsed.materials` to `setScene`.
- Tests are `bun:test` (`bun test src/validation`: 8 pass); `tsc --noEmit` and `biome check` clean; no lockfile changes.

The review's other points (vitest→bun:test in the page test, byte-accurate size cap, capture-protocol seam) are being addressed on #720.

---

> [!NOTE]
> **Low Risk**
> Scoped to build JSON validation and editor save/import; invalid materials are warned and skipped without blocking import.
> 
> **Overview**
> **Load Build** no longer drops the top-level `materials` table, which fixes custom finishes silently reverting when nodes reference `scene:<id>` slot refs.
> 
> `validateBuildJson` now accepts optional `materials`, validates each entry with `SceneMaterial.safeParse` (normalized output for the scene store), skips bad entries with `invalid_materials` warnings that name the ids, and includes valid materials on `ParsedBuildJson`. The editor **Save Build** export includes scene `materials`, and **Load Build** passes `parsed.materials` into `setScene` on confirm. Tests cover valid carry-through, partial invalid entries, and non-object `materials`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2996a1190dadb9b7e31e084463758598827afc05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/validation/validate-build-json.test.ts
+++ b/packages/core/src/validation/validate-build-json.test.ts
@@ -126,3 +126,48 @@ describe('validateBuildJson with registered plugin kinds', () => {
     expect(result.schemaIssues[0]?.nodeType).toBe('trees:tree')
   })
 })
+
+describe('scene materials', () => {
+  const minimalGraph = () => ({
+    nodes: {
+      building_1: { id: 'building_1', type: 'building', children: ['level_1'] },
+      level_1: { id: 'level_1', type: 'level', children: [] },
+    },
+    rootNodeIds: ['building_1'],
+  })
+
+  test('carries valid materials through to parsed', () => {
+    const result = validateBuildJson({
+      ...minimalGraph(),
+      materials: {
+        mat_a: {
+          id: 'mat_a',
+          name: 'Measured cabinet',
+          material: { properties: { color: '#595c5a' } },
+        },
+      },
+    })
+    expect(result.ok).toBe(true)
+    expect(result.parsed?.materials?.mat_a?.name).toBe('Measured cabinet')
+  })
+
+  test('skips invalid material entries with a warning, keeps the rest', () => {
+    const result = validateBuildJson({
+      ...minimalGraph(),
+      materials: {
+        mat_ok: { id: 'mat_ok', name: 'Fine', material: {} },
+        mat_bad: { name: 42 },
+      },
+    })
+    expect(result.ok).toBe(true)
+    expect(Object.keys(result.parsed?.materials ?? {})).toEqual(['mat_ok'])
+    expect(result.warnings.some((w) => w.code === 'invalid_materials')).toBe(true)
+  })
+
+  test('warns when materials is not an object', () => {
+    const result = validateBuildJson({ ...minimalGraph(), materials: 'nope' })
+    expect(result.ok).toBe(true)
+    expect(result.parsed?.materials).toBeUndefined()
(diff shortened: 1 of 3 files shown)


Aymericr (changes requested):
Thanks for splitting this out — the core change is right, and I checked the resolution path: keying the kept table by record key matches how every `scene:` ref is resolved at runtime (`slot-paint.ts`, `wall/paint.ts`, etc.), and the normalization stance matches how the hosted loader already parses materials, so import parity is correct. The `handleConfirmImport` retype to `ParsedBuildJson` also finally matches `LoadBuildDialog`'s `onConfirm` signature instead of a drifting inline copy. Good.

One thing must change before merge: `handleSaveBuild` in the same file (settings-panel/index.tsx:234) still exports `{ nodes, rootNodeIds, installedPlugins }`. Paint a finish, Save Build, Load Build that file — the finish reverts to default, which is exactly the bug in your description, still alive in the round-trip through this panel. As it stands the fix only helps files that already carry `materials` (API exports, hand-edited). Subscribe `materials` from the store and add it to `sceneData`; that completes the loop, and a Save→Load round-trip is the natural test for it.

Two smaller notes:

- The `invalid_materials` warning is count-only. Since the audience is hand-edited files, append the skipped ids to the message so the file is repairable — the node-level warnings in this validator all name their node.
- `collections` have the identical gap on both sides (Save omits them, `validateBuildJson` drops them, confirm doesn't pass them — while `setScene` accepts `extra.collections` and the hosted loader round-trips them). Don't fix it here; I'll file it as a follow-up unless you want to take it in a separate PR.

With the Save Build line added this is ready. (CI run is approved.)

alxbouchard:
Done in 2996a11 — Save Build now subscribes `materials` from the store and exports it, closing the round-trip (paint a finish → Save → Load keeps it). The `invalid_materials` warning now names the skipped ids. Agreed on leaving `collections` to your follow-up.

Aymericr (approved):
Verified the follow-up commit: Save Build now exports the materials table (with the why captured in the comment), and the invalid_materials warning names the skipped ids with a test pinning it. The round-trip is closed. Merging.

PR #766, opened 2026-09-06
fix(editor): restore scene auto-framing on load

ActArtech:
## What does this PR do?

Restores scene auto-framing on load and stops the level-follow effect from clobbering the framed camera pose.

- Mount `useAutoFrame()` in `EditorContent` again (it was dropped in an earlier refactor)
- On viewer scene-ready, re-emit `camera-controls:fit-scene` so fast client navigations keep the framed pose
- Level-follow: skip the initial null→level transition; only apply the default look-at when the scene graph is empty
- Expose `window.__pascalCameraControls` without a `NODE_ENV` gate (client Turbopack leaves `process` undefined)

## How to test

1. `bun test packages/editor/src/hooks/use-auto-frame.wiring.test.ts` — all pass
2. `bun run check packages/editor/src/hooks/use-auto-frame.wiring.test.ts packages/editor/src/components/editor/custom-camera-controls.tsx packages/editor/src/components/editor/index.tsx`
3. Optional smoke with `bun dev`: open a non-empty scene and confirm the camera frames the content instead of a black / default distant pose

N/A for the unit path (wiring + controls). Optional recording of load framing welcome if reviewers want visual proof.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes load-time camera orchestration and level-follow timing; regressions would show wrong framing or fighting camera poses, but no security or data paths are touched.
> 
> **Overview**
> **Restores automatic orbit framing when a scene loads** by wiring `useAutoFrame()` back into the editor and emitting `camera-controls:fit-scene` again once the viewer reports scene-ready (using XZ bounds), so fast client navigations are not overridden by other camera effects.
> 
> **Refactors fit-scene handling** into `fit-scene-framing.ts` (`computeFitSceneLookAt`, queue/apply/ignore plans). `CustomCameraControls` queues pending fits while first-person mode or an unmounted orbit ref is active, then flushes after orbit resumes (including `requestAnimationFrame` retries).
> 
> **Adjusts level-follow camera behavior** so the default `(20,20,20)` look-at runs only on first load when the scene graph is empty; vertical pans happen only on real level or `levelMode` changes, not on every initial selection path.
> 
> **Exposes `window.__pascalCameraControls` unconditionally** (drops `NODE_ENV` gating that broke under Turbopack). Adds unit tests for framing logic and source wiring assertions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7adddc0bd42fbf42afac955e701411a3c417f3c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/editor/src/components/editor/custom-camera-controls.tsx
+++ b/packages/editor/src/components/editor/custom-camera-controls.tsx
@@ -529,7 +529,6 @@ export const CustomCameraControls = ({ paused = false }: { paused?: boolean }) =
 
   const previousLevelIdRef = useRef<AnyNodeId | null>(null)
   const previousLevelModeRef = useRef(levelMode)
-  const skippedInitialLevelSelectRef = useRef(false)
   useEffect(() => {
     if (isPreviewMode || isFirstPersonMode || isRestoringFirstPersonPose()) return
     const previousLevelId = previousLevelIdRef.current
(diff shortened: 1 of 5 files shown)


Aymericr (changes requested):
There is still a blocking lifecycle case in the current head: if the scene becomes ready while first-person mode is active, the new `fit-scene` emit is ignored, and the level-follow effect returns before consuming its first-load state. When orbit controls resume, there is no later guaranteed fit, so a populated scene can remain on the default/unframed camera. The source-string test does not exercise this state transition.

Please cover scene-ready-during-first-person followed by returning to orbit with a behavioral test, and make the pending frame survive until orbit controls can apply it. I’m marking this for human product review as well because this changes load framing, level-follow, and the production exposure of `window.__pascalCameraControls` together.

ActArtech:
Thanks for the review, Aymericr, really helpful catch.

Addressed the first-person / scene-ready case: `fit-scene` is now queued while orbit is unmounted and applied when orbit remounts. Added a behavioral test for scene-ready-during-FP → return to orbit → pending frame applies.

Rebased onto latest `main`. Ready for another look when you have a moment.

PR #769, opened 2026-09-07
editor: Add duct and pipe fittings, routing, and system checks

sudhir9297:
## What does this PR do?

- **Surface-aware duct and pipe routing**
  - Issue: Runs could fall through walls or ceilings, snap to the wrong level, or stay trapped on a wall after routing moved into free space.
  - Fixed: Duct and DWV drawing now follows walls, floor and roof surfaces, and ceiling undersides before falling back to the active-level grid.
  - Grid, angle, magnetic, and spacing controls remain available while drawing in both views.

- **Expanded fitting and accessory catalogs**
  - Issue: The catalog exposed only a small fitting set and grouped unrelated MEP choices together.
  - Fixed: Duct now includes elbows, tees, crosses, reducers, transitions, end caps, dampers, access panels, and couplings; pipe includes elbows, wyes, sanitary tees, crosses, end caps, cleanouts, reducers, and couplings.
  - Each choice has its own thumbnail, and opening MEP starts with no tool selected so only the chosen system's catalog is shown.

- **Shape-aware duct adapters**
  - Issue: Connecting different duct profiles could create a same-shape reducer or leave rectangular and round runs joined without a visible adapter.
  - Fixed: Size changes use reducers, while round, rectangular, and oval shape changes create and orient transitions automatically.

- **Compatible and durable connections**
  - Issue: Proximity alone could join incompatible systems, multi-floor DWV runs could lose connectivity, and deleting a run could leave damaged fitting topology.
  - Fixed: Connections now compare typed systems in world space, show compatibility feedback, preserve vertical connections across floors, and simplify or remove affected fittings when runs are deleted.

- **Connected-run editing**
  - Issue: Continuations, branches, crossings, endpoint moves, and whole-run moves could produce inconsistent previews or many history updates during one gesture.
  - Fixed: Shared continuation and branch planning keeps 2D and 3D behavior aligned, previews edits through live overrides, and commits each placed or edited item as one undo step.

- **DWV slope behavior**
  - Issue: Drainage grade was hard to apply, and the full slope editor made the in-canvas routing card too crowded.
  - Fixed: Waste presets and the `S` shortcut apply the standard grade without showing the large slope form or live readout over the drawing.
  - The side inspector keeps drainage checks for flat, steep, or overlong trap-arm conditions.

- **System checks**
  - Issue: Open ends, disconnected branches, incompatible joins, unsupported hangers, and likely intersections were difficult to find from the editor.
  - Fixed: A system-check panel lists these findings and lets the reviewer select and reveal each affected part.

- **Individual hanger control**
  - Issue: Automatic hangers could not be adjusted when a generated support was unsuitable or missing.
  - Fixed: Each hanger can be moved, skipped, or assigned to a support host, with a visible warning when no support is available.

## How to test

1. Run `bun run checks`, `bun run lint`, and `bun run build`.
   - Formatting, lint, type checks, package builds, and both editor production builds should pass.

2. Run `bun --cwd packages/nodes test`.
   - The node suite should report 2,069 passing tests and one skipped test.

3. Start the editor, open MEP, and choose Duct or DWV Pipe.
   - No MEP item should be selected initially, unselected tiles should be grey, and the catalog should show only choices for the selected system.

4. Draw on a wall, move into free space, return to the wall, and repeat on a ceiling underside, floor, roof, and another level.
   - The run should stay on the hovered surface, return to surface snapping correctly, and avoid snapping to a transparent ceiling top face.

5. Change the active snap mode and snap spacing while drawing in both 2D and 3D.
   - The preview and committed run should immediately use the selected mode and spacing.

6. Connect rectangular, round, and oval duct runs with matching and different sizes.
   - Same-shape size changes should create reducers; shape changes should create correctly rotated transitions between both profiles.

7. Draw a vertical DWV stack across floors, add branches, and toggle the standard waste grade with `S`.
   - Connections should remain part of the same compatible system, and the system checks should report any invalid drainage conditions.

8. Select a duct or pipe run, move an endpoint or the whole run, then use undo and redo.
   - The live preview should stay connected, and the completed gesture should create one history entry.

9. Open the system checks and hanger controls for a run.
   - Findings should reveal their affected parts, and individual hangers should support move, skip, host selection, and missing-support warnings.

Not added yet.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Touches core scene deletion hooks, distribution system graph connectivity, and pointer/grid event plumbing used by many tools; schema version bumps on duct nodes add migration surface.
> 
> **Overview**
> **MEP build UI** now opens the MEP group without auto-arming duct draw; duct and pipe fitting types are picked through **`ToolOptionsPanel`** with catalog thumbnails instead of one-off “Add Fitting” buttons.
> 
> **Routing and placement** gain architectural surface raycasting on **`GridEvent`** (walls, ceilings, slabs, roofs), ceiling-underside picking, and optional placement lattice anchors so wall-mounted runs and fitting ghosts snap on the hovered host—not the floor grid.
> 
> **Duct and DWV models** expand fitting enums (caps, dampers, access panels, couplings, cleanouts, etc.), add **auto hangers** / overrides, **wall attachment** metadata, and richer reducer/transition geometry with **`inletShape` / `outletShape`**. Runs support **continue/branch** affordances in plan (elbow→tee / tee→cross promotion) and drafting opts into **`cancelOnHistoryJump`** so undo/redo while drawing cancels in-flight drafts.
> 
> **System connectivity** joins ports in **world space** (level elevation + building transform), ignores mismatched **`system`** tags at coincident joints, and exports **`collectSystemPorts`** / **`distributionPointToWorld`**. Node **`onDelete` / `onDeleteCascade`** now receive **`requestedDeleteIds`** vs the expanded pending set for smarter fitting cleanup.
> 
> Smaller editor fixes: lazy custom inspector panels receive the live **`node`**, dev bootstrap re-registers builtin kinds on HMR, and floorplan midpoint handles can use **`activation: 'action'`** (click-only).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d1bd7fe4b5680d99ed71ac768d9ca1a5d5d49e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/editor/src/hooks/use-grid-events.ts
```diff
@@ -4,25 +4,46 @@ import {
...
  */
 export function useGridEvents(gridY: number) {
   const { camera, gl } = useThree()
+  const interactionScope = useInteractionScope((state) => state.scope)
+  const semanticSurfaceQueryRef = useRef(false)
+  semanticSurfaceQueryRef.current =
+    interactionScope.kind === 'placing' ||
+    interactionScope.kind === 'moving' ||
+    (interactionScope.kind === 'drafting' &&
+      (interactionScope.tool === 'duct-segment' || interactionScope.tool === 'pipe-segment'))
```
wass08: This hardcoded kind list appears 4× in this file (lines 30, 67, 137, 190). Framework code shouldn't branch on specific node kinds — resolve this from the registry instead, e.g. a `drafting.surfaceQuery` flag on the `NodeDefinition` looked up via `nodeRegistry.get(scope.tool)`. As written, `lineset` / `liquid-line` drafting silently misses the surface-query path.

packages/editor/src/lib/history.ts
```diff
@@ -72,7 +73,15 @@ function refreshSceneAfterHistoryJump() {
   }
 }
 
+export function isRunDrafting(): boolean {
+  const scope = useInteractionScope.getState().scope
+  return (
+    scope.kind === 'drafting' && (scope.tool === 'duct-segment' || scope.tool === 'pipe-segment')
```
wass08: Same kind list as `use-grid-events.ts` — worth deriving both from one registry-driven helper (e.g. `def.drafting.cancelOnHistoryJump`) so the next run-drafting kind gets this behavior by declaration instead of by editing framework files.

wass08 (changes requested):
Nice work overall — the Shift-bypass migration in the MEP move/selection tools is exactly the cleanup `interaction-scope.md` asks for, schema changes are all backward-compatible, and the new hanger/fitting behavior hangs off `NodeDefinition` the way it should.

One blocker before merge:

**Kind-specific branching landed in framework code.** This PR introduces the predicate `scope.kind === 'drafting' && (scope.tool === 'duct-segment' || scope.tool === 'pipe-segment')` in five places across `packages/editor` — four in `hooks/use-grid-events.ts` (lines 30, 67, 137, 190) and one in `lib/history.ts` (`isRunDrafting`, line 79). Per `wiki/architecture/` (layers + node-definitions), new kind-specific dispatch in a framework package is a regression: the behavior belongs on the kind's `NodeDefinition`, resolved through `nodeRegistry`.

It's also a real maintenance trap, not just style: `lineset` and `liquid-line` draft runs too, but aren't in the list — so they silently don't get semantic surface queries or the undo-cancels-draft behavior, and the next MEP-ish kind has to find and extend a hardcoded list in two files.

Suggested fix (mechanical): declare it on the definition, e.g.

```ts
// NodeDefinition
drafting: {
  surfaceQuery: true,        // use-grid-events: architectural surface raycast + level frame
  cancelOnHistoryJump: true, // history: undo/redo cancels the in-flight draft
}
```

then replace the five predicates with one helper that resolves `nodeRegistry.get(scope.tool)?.drafting` (or equivalent capability flags). No behavior change for duct/pipe, and the other run kinds can opt in by declaration.

Everything else I looked at checked out clean — happy to approve once this is registry-driven.

PR #833, opened 2026-09-10
perf(core): guard bulk slab changes in the spatial-grid subscriber

RIRuhAI:
## What does this PR do?

The spatial-grid subscriber runs `markNodesOverlappingSlab` for every slab a store write adds, removes or reshapes: two full node scans per slab, so a scene load, reload, paste or import is O(slabs × nodes). On a 4,645-slab scene (7,143 nodes) that scan alone blocked the main thread for ~5 s on every load — a real-Chrome CPU profile showed 5.9 s inclusive inside this scan out of a 7 s `setScene` block, enough for Chrome's "Page Unresponsive" dialog on a loaded machine.

When one write changes 32 or more slabs, this dirties every node any slab could affect once — walls, ceilings, stairs and level-hosted floor-placed kinds, a superset of the per-slab marks — and skips the per-slab scans. `setScene` marks every node dirty right after `set()` regardless, so load-time behaviour is unchanged; single-slab edits keep the targeted rules (`markSlabChangeDependents`, covering-below, deck stairs).

**Measurement caveat:** the numbers were taken with an equivalent patch applied to the published `@pascal-app/core@1.0.0-beta.5` (before #805), not re-taken on this source. Same mechanism. Max main-thread freeze on that scene went 6.4 s → 0.4–0.9 s; time to a ready model 31.7 s → ~21 s. Happy to re-run on a fixture you prefer.

## How to test

1. `bun test --cwd packages/core src/hooks/spatial-grid/spatial-grid-sync.test.ts` — new `bulk slab-change guard` block: 31 added interior plates leave a non-overlapping wall clean (per-slab path); 32 mark it (superset path); a 3,000-slab write stays linear (well under 1.5 s, versus seconds unguarded); later single-slab edits stay targeted.
2. Load a scene with hundreds of slab nodes with `?perf` and compare the `setScene` span before and after.
3. Full core suite passes (1,490 tests); `tsc --build` and `bun check` clean.

No visual change. Numbers above; a before/after CPU profile of the scan is available on request.

## Checklist

- [ ] I've tested this locally with `bun dev` — verified through the package test suites, `tsc --build` and `bun check` instead; no editor session was run for this change

---

> [!NOTE]
> **Medium Risk**
> Changes when and which nodes are marked dirty in the spatial-grid subscriber; bulk mode trades precise overlap scans for a fixed superset sweep, so correctness depends on that superset matching real slab dependents outside full-scene hydration.
> 
> **Overview**
> Fixes main-thread freezes on scene load, paste, and import when thousands of slabs change in one store write. The spatial-grid subscriber used to run **two full node scans per added, removed, or reshaped slab**; this adds a **bulk guard** at **32+ slab changes** that calls **`markAllSlabDependents`** once (walls, ceilings, stairs, level-hosted floor-placed nodes) and **skips** the per-slab overlap and **`markSlabChangeDependents`** paths.
> 
> Exports **`BULK_SLAB_CHANGE_THRESHOLD`**, **`countBulkSlabChanges`**, and **`markAllSlabDependents`**. Dirty marks on bulk writes can be a **superset** of the old per-slab marks (e.g. a non-overlapping wall may be marked at the threshold); **single-slab edits** still use the targeted rules. Tests cover counting, threshold behavior, a 3,000-slab performance bound, and that later one-slab edits stay precise.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7997a2f76508d656f95025b7b7b8482075f267b7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/core/src/hooks/spatial-grid/spatial-grid-sync.test.ts
+++ b/packages/core/src/hooks/spatial-grid/spatial-grid-sync.test.ts
@@ -9,6 +9,8 @@ import useLiveTerrain from '../../store/use-live-terrain'
 import useScene, { clearSceneHistory } from '../../store/use-scene'
 import { spatialGridManager } from './spatial-grid-manager'
 import {
+  BULK_SLAB_CHANGE_THRESHOLD,
+  countBulkSlabChanges,
   initSpatialGridSync,
   markCoveringDependentsBelow,
   markLevelHeightDependents,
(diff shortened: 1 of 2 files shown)


Aymericr (approved):
The algorithmic guard is sound: the bulk path marks a true superset of every per-slab consumer, while single-slab writes retain the precise overlap/deck/covering rules. I also merged the head onto current main locally and verified the focused spatial-grid suite (24 tests / 64 assertions), the core build, and Biome. The 3,000-slab regression completed in 182 ms here.

PR #845, opened 2026-09-11
feat(cli): ship a small CLI that downloads the web editor runtime

Aymericr:
## What does this PR do?

Splits `@pascal-app/cli` into a small npm package plus a web-runtime archive that is downloaded on first need, and makes MCP-only mode work with no web runtime at all.

Today the published CLI is **65 MB compressed / 102 MB unpacked** because `stage-runtime` copies the standalone Next editor app into `dist/runtime`. The CLI code itself is 0.1 MB. Every agent host that only wants `pascal mcp connect` pays for a Next server it never starts.

**After this PR: 0.46 MB compressed / 2.46 MB unpacked / 68 files.**

### 1. MCP-only mode needs no web runtime

`pascal mcp connect` now starts (or attaches to) the managed MCP service on its own. The new `src/mcp-service.ts` owns the service lifecycle — spawn, loopback bearer token, `run/mcp.json` state, health check, reuse, stop — which `editor-process.ts` previously did as a child of the editor process. `connectManagedMcp` no longer calls `startEditor()` and no longer requires `editorHealthy`. `editorUrl` values keep their existing `PASCAL_EDITOR_ORIGIN` fallback, and starting the editor later repoints the running service at the real origin (restart + repoint) rather than leaving a stale one.

`src/process-control.ts` holds the pid/signal helpers both lifecycles now share.

### 2. The package is split

`dist/` ships CLI code, `services/pascal-mcp.mjs` (2.3 MB, unchanged), and a new `dist/runtime-source.json`:

```json
{
  "version": "1.0.0-beta.1",
  "url": "https://github.com/pascalorg/editor/releases/download/@pascal-app/cli@1.0.0-beta.1/pascal-web-runtime-1.0.0-beta.1.tar.gz",
  "sha256": "b92fe413e519c46af559e4b9c41baaf17c3a70dd8df23b6099fcac0c0228b5b9",
  "size": 67340245
}
```

### 3. Download on first need

`ensureWebRuntime` (`src/runtime-download.ts`) resolves, in order:

1. `--runtime <path|archive>` (new flag on `editor`, `start`, `update`, `install`)
2. `PASCAL_BUNDLED_RUNTIME_DIR` (existing)
3. an already-installed runtime for this exact CLI version under the data dir
4. download `runtime-source.json.url` with streamed progress, verify SHA-256, extract, install

Verification is **fail-closed**: a mismatch throws `runtime_digest_mismatch` and deletes the downloaded temp file (a caller-supplied archive is left alone). Everything then hands off to the untouched `installBundledRuntime` seam, so the atomic temp-dir install, manifest validation, `.install-*` reaping and `runtime-install.lock` are reused verbatim — concurrent first runs serialise on the existing lock.

`src/http-download.ts` does the transfer on `node:https` with no new dependency: manual redirect following, https-only, `HTTPS_PROXY`/`ALL_PROXY`/`NO_PROXY` (including `*`, `.suffix`, `host:port` forms) through a `CONNECT` tunnel agent. On network failure the error names the archive URL, the expected digest, `pascal editor --runtime <archive>` and `HTTPS_PROXY`.

`src/tar.ts` is a small deterministic tar writer/reader: sorted names, zeroed mtime/uid/gid, ustar + GNU long-name entries, gzip level 9. Extraction rejects path traversal, absolute and drive-letter paths, symlinks, hard links and bad header checksums with `invalid_runtime_archive`.

### 4. Build and release plumbing

`stage-runtime` now also writes `build/pascal-web-runtime-<version>.tar.gz` + `.sha256` and `dist/runtime-source.json`. Two independent runs produced the identical digest above.

Release ordering in `release.yml`:

1. `Validate portable editor runtime` builds the editor and stages the archive (unchanged step, new side effects).
2. `Publish CLI` **guards before publishing**: the archive, its `.sha256`, `dist/services/pascal-mcp.mjs` and a `dist/runtime-source.json` whose `version` equals `$CLI_VERSION` must all exist. Then `npm publish`.
3. `Commit version bumps & tag` pushes `main` + the `@pascal-app/cli@<version>` tag atomically.
4. A new `Upload the CLI web runtime release asset` step creates the GitHub release for that tag if absent (`--prerelease` on the beta dist-tag) and `gh release upload --clobber`s the archive + digest.

The asset upload has to follow the tag push (the release hangs off the tag), so there is a residual window of roughly 30–60 s between `npm publish` and the asset being live during which a brand-new install would fail its download. The pre-publish guards make that window a scheduling artefact rather than a correctness one, and `gh release upload --clobber` makes re-running the job idempotent. Worth a follow-up if we ever want it closed properly (publish to a staging dist-tag, upload, then retag).

### 5. Budgets

`smoke-packed-runtime.ts` now checks npm package ≤ 3 MB compressed / ≤ 10 MB unpacked / ≤ 250 files and the archive ≤ 70 MB. Measured: **0.46 MB / 2.46 MB / 68 files** and a **64.2 MB** archive (106 MB installed). Archive headroom is only ~8%, so a materially larger editor build will trip the budget deliberately.

### 6. Docs and tests

`packages/cli/README.md` gains a `## The web editor runtime` section (what downloads when, sizes, the resolution order, digest verification, the offline `curl` + `--runtime` recipe, proxy support, and that agent tools need none of it). `SETUP.md`, the root `README.md` and both skill `references/setup.md` local sections say `pascal mcp connect` needs no runtime download; the skill statements are version-qualified so they do not contradict the currently published `1.0.0-beta.1`. Frontmatter untouched, bundles still standalone, `bun run skills:validate` passes. `CHANGELOG.md` has one Unreleased line.

38 new unit tests across `tar.test.ts` (9), `runtime-download.test.ts` (22) and `mcp-service.test.ts` (7) cover digest verification, the resolution order, the `runtime-source.json` validation table, proxy resolution, extraction safety, and MCP-without-a-runtime including concurrent starts and origin repointing. The download tests point at an unroutable `https://127.0.0.1:1`, so the suite stays offline.

## How to test

1. Build the editor once and stage the split artifacts:
   ```bash
   PASCAL_PORTABLE_BUILD=1 bun run build --filter editor
   cd packages/cli && bun run build && bun run stage-runtime
   ```
2. Run the smoke test — phase 1 is MCP-only with no runtime present, phase 2 tampers a byte in the archive, phase 3 installs it through `--runtime`:
   ```bash
   bun run smoke-runtime
   ```
   ```
   MCP-only mode exposed 46 tools and stored a scene with no web runtime installed.
   Tampered archive rejected: The Pascal web runtime archive does not match the digest published with this CLI.
   Editor installed from pascal-web-runtime-1.0.0-beta.1.tar.gz on port 56428, MCP on port 56431, and a scene round-tripped between MCP and the CLI.
   Packed CLI smoke passed (0.5 MB compressed, 2.5 MB unpacked, 68 files).
   Web runtime archive pascal-web-runtime-1.0.0-beta.1.tar.gz (64.2 MB) verified against https://github.com/pascalorg/editor/releases/download/@pascal-app/cli@1.0.0-beta.1/pascal-web-runtime-1.0.0-beta.1.tar.gz
   ```
3. `cd packages/cli && bun test` → 102 pass, 0 fail, 288 expect() calls across 11 files.
4. `bun run check` (clean, 2142 files), `bun run check-types` (clean), `bun run skills:validate` (2 skills + 51 policy tests).
5. Confirm the package size yourself: `cd packages/cli && npm pack --dry-run --json` → `size 487831`, `unpackedSize 2583923`, `entryCount 68`.

Reviewers who want to exercise the real download path need the release asset to exist, which only happens on the next CLI release; until then `--runtime build/pascal-web-runtime-<version>.tar.gz` covers the same code path minus the transfer.

Not applicable — CLI and release plumbing only, no visual surface.

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes how users install and verify the editor (download, digest, release asset ordering) and splits MCP vs editor lifecycle; strong tests and fail-closed verification, but a brief npm-publish-before-asset window remains.
> 
> **Overview**
> **`@pascal-app/cli` shrinks to a ~0.5 MB npm package** that ships the CLI and bundled MCP service; the Next.js web editor is published per version as a **GitHub release archive** referenced by `dist/runtime-source.json` (URL, size, SHA-256).
> 
> **Editor commands** (`editor`, `start`, `update`, etc.) gain **`--runtime`** and call **`ensureWebRuntime`**: resolve an override or installed copy, or **download once**, **verify digest fail-closed**, extract via new **`tar.ts`**, then reuse existing install/lock logic. **`http-download.ts`** handles HTTPS-only transfers with **`HTTPS_PROXY` / `NO_PROXY`**.
> 
> **MCP is decoupled from the web runtime**: **`mcp-service.ts`** manages its own process and `run/mcp.json`; **`pascal mcp connect`** no longer starts the editor. Editor start still brings MCP up and can **repoint** `PASCAL_EDITOR_ORIGIN` when the editor port changes.
> 
> **Release workflow** requires the staged archive, digest, MCP bundle, and matching `runtime-source.json` **before** `npm publish`, then **uploads the runtime tarball** to the `@pascal-app/cli@<version>` GitHub release after the tag is pushed. **`stage-runtime`** writes `build/pascal-web-runtime-*.tar.gz` and drops MCP from the runtime tree (manifest **schema v2**). Docs, changelog, and expanded smoke/unit tests reflect the split and offline path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6abc4ed329fdb5dc4a776a1fad681a29b45aa442. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Aymericr:
CI ran the new smoke phases on `macos-latest` in ~2 min:

```
Staged Pascal web runtime 1.0.0-beta.1 at .../packages/cli/build/runtime
Archived 3309 entries to .../pascal-web-runtime-1.0.0-beta.1.tar.gz (43.16 MB)
MCP-only mode exposed 46 tools and stored a scene with no web runtime installed.
Tampered archive rejected: The Pascal web runtime archive does not match the digest published with this CLI.
Editor installed from pascal-web-runtime-1.0.0-beta.1.tar.gz on port 49487, MCP on port 49490, and a scene round-tripped between MCP and the CLI.
Packed CLI smoke passed (0.5 MB compressed, 2.5 MB unpacked, 68 files).
```

Two notes on the archive size. CI produced **43.2 MB / 3309 entries** where my local run produced **64.2 MB / 3312 entries** — the difference is what the standalone Next build traced out of each machine's `node_modules` (a full local install hoists more than `--frozen-lockfile` does), so the 64.2 MB figure in the description is the pessimistic one and real headroom under the 70 MB budget is larger than it looks.

That also means the digest is reproducible for a given tree on a given machine, not across machines. It does not need to be: `stage-runtime` writes the archive and `dist/runtime-source.json` in the same job that publishes, so the embedded digest always describes the asset that job uploads.

PR #854, opened 2026-09-11
fix(editor): clean up orphaned local asset:// files from IndexedDB

### LINKED ISSUE #733: Local asset storage never cleans up: deleting an asset:// scan or guide orphans its File in IndexedDB
Split out of #726 (review note) so it doesn't get lost.

`handleDelete` in `packages/editor/src/components/ui/sidebar/panels/site-panel/index.tsx` only routes `http(s)` URLs to `onDeleteAsset`. Assets persisted locally through `saveAsset` (`packages/core/src/lib/asset-storage.ts`, `asset://<id>` URLs) are never removed from IndexedDB when the referencing node is deleted.

Guide images have had this gap since local guide storage landed; with #726 the same path now stores scans, which are up to 200 MB each, so the orphaned data is no longer trivia.

Two layers to it:

1. **Immediate**: when a node holding the last reference to an `asset://` URL is deleted, delete the IDB entry (mind undo — a deleted node can come back with history, so eager deletion needs either a grace period or resurrection re-save).
2. **Structural**: `loadAssetUrl`'s module-level `urlCache` never revokes object URLs either. A small ref-counted asset lifecycle in `asset-storage.ts` would cover both.

An interim mitigation could be a sweep on scene load: drop IDB entries whose ids appear in no loaded node.

Zhao0335:
﻿## What does this PR do?

Fixes #733.

Deleting a local scan or guide node only routed `http(s)` URLs to `onDeleteAsset`. Local `asset://` Files — including scans up to ~200 MB — stayed in IndexedDB forever after the referencing node was gone.

This PR schedules a **delayed delete of the specific File** when the last live node that pointed at it is removed:

- Scene-panel delete of a scan/guide schedules the local File for removal after a 120s grace period (undo can still restore the node and load the blob).
- Reference-panel delete and replace paths do the same for guide images.
- `deleteAsset` in `@pascal-app/core` removes the IndexedDB entry and revokes any cached object URL.
- A **scene epoch** is bumped on every `applySceneGraphToEditor`. Timers scheduled under a previous epoch refuse to delete, so switching projects cannot wipe a File another saved scene still uses.

It does **not** sweep "all unreferenced assets" on scene load. IndexedDB is origin-global and not scoped by scene.

## How to test

1. Self-host the editor (`bun dev`), open a project, and upload a local guide image.
2. Delete the node from the Scene panel (or the reference panel).
3. Within ~2 minutes, Ctrl+Z the delete — the guide should still resolve.
4. Let the grace period elapse (or wait for the timer) — the IndexedDB entry for that asset should be gone (`Application -> IndexedDB` in DevTools).
5. Delete a local asset, then immediately open another project before the timer fires — the File must remain loadable when you reopen the first scene.
6. Open another scene that still has local assets — they must remain loadable (no cross-scene wipe).

N/A — storage lifecycle only, no visual change.

## Checklist

## Validation

- `packages/core`: 1532 tests pass (`bun run test`)
- `packages/editor`: `local-asset-lifecycle.test.ts` 6/6 pass
- `@pascal-app/core` `tsc --build` clean
- `@pascal-app/editor` `check-types` clean

## Review follow-ups

1. Removed the fire-and-forget origin-wide orphan sweep so opening one scene cannot delete assets referenced by other graphs.
2. Added a scene epoch: delayed deletes scheduled under a previous `applySceneGraphToEditor` epoch are skipped, so a project switch during the grace window cannot delete a File another saved scene still references.

---

> [!NOTE]
> **Medium Risk**
> Changes how local blobs are deleted across origin-global IndexedDB; incorrect keep-sets or calling GC with a partial scene inventory could remove files still referenced by other scenes.
> 
> **Overview**
> Adds **safe, opt-in garbage collection** for local `asset://` blobs in IndexedDB (#733), replacing risky origin-wide or mount-time sweeps.
> 
> **`@pascal-app/core`** grows `deleteAsset`, `listLocalAssetUrls`, and `sweepLocalAssetsExcept`, plus deep collectors (`collectNodeAssetUrls`, `collectSceneAssetUrls`, `collectGraphAssetUrlsFromParts`) that walk nodes and materials for `asset://` strings without deleting anything.
> 
> **`apps/editor/lib/local-asset-gc.ts`** implements `runLocalAssetGc`: it builds a keep-set from the live graph, `localStorage` draft scene, and every server scene (via `/api/scenes`), then sweeps only when that inventory is provably complete—**returns `null` and skips** if listing fails, a per-scene fetch fails, or the scene list hits the 500 cap. Live nodes are re-read immediately before sweeping.
> 
> Editor UI (**reference panel**, **site panel**) stops implying panel-level IndexedDB cleanup; comments state removal/replace flows defer to core scene mutations. A stale bootstrap comment is removed from `scene-loader.tsx`. Tests cover collectors, sweep behavior, and GC abort paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 276e5b0de6407b5b5210c2984e42c2ea39f388e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Zhao0335:
Addressed the Bugbot finding in d44b09d: removed the scene-load orphan sweep. IndexedDB is origin-global, so sweeping entries missing from the just-loaded graph could wipe assets still referenced by other scenes. Cleanup is now only the undo-safe delayed delete of a specific sset:// File after its last live node is removed.

Zhao0335:
Addressed the latest Bugbot finding: delayed deletes now capture a scene epoch at schedule time. pplySceneGraphToEditor bumps the epoch; timers from a previous epoch refuse to deleteAsset. Switching projects during the grace window no longer deletes origin-global IndexedDB Files that another saved scene still references. Covered by a new unit test.

Zhao0335:
This finding is already fixed on the current HEAD (cb97731), which is ahead of the reviewed 7ae66a2.

- pplySceneGraphToEditor bumps a scene epoch (umpLocalAssetSceneEpoch).
- scheduleLocalAssetDelete captures that epoch at schedule time; the timer returns without deleteAsset if the epoch changed (project/scene switch).
- Covered by unit test: does not delete after a scene switch (another graph may still use the File).

No further code change needed for this comment.

Aymericr (changes requested):
Blocker: cleanup is attached to two UI handlers rather than the deletion lifecycle, so the bug still reproduces through supported delete paths.

For example, `packages/editor/src/hooks/use-keyboard.ts` deletes a selected guide/scan directly with `useScene.getState().deleteNode(...)`, and the generic selection/delete surfaces in `selection-routing.ts` and `group-actions.ts` also bypass `scheduleLocalAssetDelete`. Deleting the same local reference with the Delete key therefore still leaves its `asset://` File in IndexedDB.

Please move the scheduling to a single path that observes all committed node removals/replacements, or route every supported guide/scan deletion path through one shared helper. Add a regression test for keyboard deletion (and ideally delete-mode/generic selection deletion) so this cannot remain UI-entry-point dependent. The 120-second grace period, live-reference recheck, cache revocation, and scene-epoch guard are otherwise sensible.

Zhao0335:
Addressed the deletion-lifecycle blocker in \$(git rev-parse --short HEAD)\:

- Local sset:// scheduling now lives in @pascal-app/core and runs from **deleteNodesAction** (and pplyNodeChanges deletes/url replacements), not the Site/reference panel handlers.
- Covers keyboard Delete (use-keyboard.ts → deleteNode), selection delete, MCP delete_node, and group actions.
- Regression test: packages/core/src/lib/local-asset-lifecycle-delete.test.ts — keyboard-style deleteNode schedules the File; a second live reference keeps it.

Grace period, live-reference recheck, cache revocation, and scene-epoch guard are unchanged.

Aymericr (changes requested):
The central delete lifecycle now covers the missing keyboard/generic paths, but the current head still has three correctness blockers before cleanup is safe: (1) URL replacement schedules deletion without a fire-time scan of all live nodes, so undo or another node sharing the same asset:// URL can lose its File; (2) the reference-panel replacement schedules before updateNode commits, so a no-op/throw can delete the still-current asset; and (3) the scene epoch only advances in applySceneGraphToEditor, while unload/import/reset paths can replace the graph through setScene/clearScene and leave an old timer valid. Please make the timer predicate query the current graph for the URL at fire time, schedule replacements only after a committed change, and put epoch invalidation at the shared graph-replacement boundary. CI also needs the three Biome organize-import fixes shown by quality. Add regressions for shared-reference replacement, undo restoration, failed/no-op replacement, and a clear/set scene transition.

Zhao0335:
Addressed the three correctness blockers and Biome organizeImports on \$head\:

1. **Replacement timers** re-query the live graph for the URL at fire time (collectSceneAssetUrls(get().nodes)), so undo or a duplicate guide sharing the same sset:// handle keeps its File.
2. **Guide replace** no longer schedules before updateNode; only the core updateNodes path schedules after the url change commits.
3. **Epoch** is bumped in unloadScene and setScene (shared graph-replacement boundary), not only pplySceneGraphToEditor.

Regressions added for shared-reference replacement, undo restoration, unreferenced replacement, and setScene epoch. packages/core: 1545 tests pass.

Aymericr (changes requested):
The new fire-time live-graph checks fix the reported undo/shared-node cases, but there is still a cross-scene data-loss blocker. `asset://` storage is origin-global, and scene/project duplication preserves those URLs. If scene A and an unopened saved scene B reference the same File, deleting the last reference in the currently loaded A schedules `deleteAsset`; staying in A for 120 seconds means the timer sees no current reference and deletes B’s File. The epoch only protects a scene switch during the grace window, not references in persisted scenes that were never loaded. This is the same ownership problem that made the original sweep unsafe.

Please do not physically delete origin-global Files based only on the active graph. The cleanup path needs a durable cross-scene reference index/refcount (updated atomically with scene persistence), per-scene asset ownership/copying, or a conservative explicit garbage-collection pass that can inspect every saved graph. Until one of those exists, this PR trades an orphan leak for silent corruption of another project. Add a regression with two persisted scenes sharing one asset URL; deleting from one must not break the other.

Zhao0335:
Addressed the cross-scene data-loss blocker in `55e04ae`.

**No longer deletes origin-global Files from the active graph.**

1. Removed automatic `deleteAsset` from `deleteNodes` / `updateNodes` timers.
2. Core now exposes explicit GC: `sweepLocalAssetsExcept(keepUrls)`.
3. Standalone editor GC (`apps/editor/lib/local-asset-gc.ts`) builds the keep-set from **current scene + localStorage scene + every `/api/scenes` graph**. If any source cannot be enumerated, it **skips** the sweep rather than partial-deleting.
4. Regression test: two persisted scenes share one `asset://` URL; sweeping with both keep-sets keeps the File.

Orphan leak may remain until a successful full enumeration runs — that is preferred over silent corruption of another project.

Zhao0335:
Fixed the two High findings on `0023ec0`:

1. **Truncated inventory** — /api/scenes has no cursor; a full page (500) is treated as incomplete and GC **skips** instead of sweeping. Listing failure still skips.
2. **In-flight uploads** — 
unLocalAssetGc takes a live-node getter and re-reads useScene **immediately before** sweepLocalAssetsExcept, and keeps initialScene urls in the extra keep-set.

Unit tests cover truncated list, failed list, and mid-GC live-graph growth.

Zhao0335:
Addressed the mount-time GC findings on `3cc6f50`:

- **Removed** automatic `runLocalAssetGc` from `SceneLoader` mount.
- `runLocalAssetGc` remains an **explicit host API only** (documented: no in-flight writes, complete inventory, spare rate budget).
- Sequential per-scene GETs no longer run on page open, so they cannot starve autosave's rate bucket.

This leaves possible orphans in IndexedDB rather than deleting Files another tab, keepalive flush, or unpersisted upload still needs.

Zhao0335:
Fixed keep-set completeness on `276e5b0`:

- `collectNodeAssetUrls` now deep-walks the node (depth 8, cycle-safe) so `item.asset.src` / `thumbnail`, `captureSession.manifestUrl`, and other nested `asset://` strings are included.
- New `collectGraphAssetUrlsFromParts` also walks `graph.materials` texture maps.
- Unit tests cover nested item fields and materials maps.

dkbbdev:
Reviewing this against #733 — verified the fix routes local `asset://` URLs into the delayed-delete scheduler so deleting a scan/guide node no longer orphans its IndexedDB File (the ~200 MB case is the one that matters on memory-constrained devices).

Questions while reading:
1. **Delete-then-recreate race** — if a user deletes a scan node and immediately re-adds a guide referencing the same underlying File before the grace period elapses, does the scheduled delete get cancelled or does it remove the File out from under the new node?
2. **Ref-counting scope** — two nodes sharing one asset URL is the obvious next failure mode; is that handled or explicitly out of scope here? If out of scope, a short note in the PR body would help future readers.
3. Minor: the scheduler timer survives page reload? (IndexedDB File pending delete + reload → does the pending entry get re-scheduled or leaked until the next delete?)

Zhao0335:
Thanks for the careful questions. The PR had changed substantially since its original description, so I corrected the title and body to match the current code.

1. **Delete then recreate:** There is no scheduled delete on node removal now. The earlier timer was removed after the cross-scene data-loss review, so recreating a reference during a grace period cannot race it.
2. **Shared URL:** There is no refcount. The explicit GC path gathers references from the live graph, the localStorage draft, and all server scenes; if the server inventory fails or is truncated, it skips deletion. `b8a6af37` also makes it retain live material textures, including those added while the inventory is running. The new regression uses a real IndexedDB File.
3. **Reload:** There is no pending timer to resume. Orphans remain until a host explicitly invokes GC when it can ensure a complete inventory and no in-flight writes.

This means the PR is groundwork for #733, not a complete automatic fix. I removed the closing `Fixes #733` language from the description and made that limitation explicit. Local targeted tests, TypeScript, Biome, and the dependency build pass; GitHub CI is running on the new head.

Zhao0335:
@Aymericr The review threads are resolved, and the latest push fixes the live-material keep-set and makes GC abort on malformed or incomplete scene inventories. All current checks pass. Could you take another look? The PR description now states clearly that this is an explicit cleanup API and does not fully close #733.

PR #882, opened 2026-09-17
fix(viewer): preserve snapshot atmosphere and capture isolation

AxiomeCG:
## Summary

- render snapshot backdrops from the active `SceneAtmosphere` source, with the scene theme as the fallback
- keep the scene MRT transparent inside the nested Three.js r186 pass so preset captures retain alpha
- make preset isolation explicitly transparent while preserving the existing host scene-slot visibility policy
- pin Environment to `1aa0bc3`, adding its global visibility control, isolation-aware runtime, and Three.js 0.186 support

## Root cause

Three.js r186 resets renderer clear state inside `RTTNode` before the nested scene pass. The snapshot MRT therefore cleared empty sky to opaque alpha even when the outer capture requested transparency. Separately, the snapshot graph always synthesized the scene-theme backdrop, so active presentation skies never reached stored thumbnails.

## Verification

- `bunx tsgo --noEmit -p packages/editor/tsconfig.json`
- `bun run check-types` — 10/10 workspace tasks passed
- `bun run build` in `packages/viewer`
- `bun test packages/editor/src/components/editor/snapshot-capture.test.ts` — 19 passed
- `bunx biome check` on every changed source and manifest
- real editor smoke at Three.js 0.186: Environment master control rendered with the documented subtext; enabling it changed the live sky, disabling it restored the native background
- real WebGPU/Metal capture QA: native and Environment skies captured correctly; preset exterior remained transparent

Environment release verification:

- `bun test` — 347 passed
- `bun run check-types`
- lab `bun run check-types`

Environment commit: https://github.com/AxiomeCG/environment/commit/1aa0bc3cbed273dbf751e9af79e446f6de105c90

## Scope correction

The `viewerSceneSlot` gate is unchanged from upstream: `{!noEditing && viewerSceneSlot}`. Neither bundled editor mount supplies this optional host slot, and Environment uses `ViewerPresentations`, not the slot. Host-only Building Lab placeholders are outside this PR. After restoring the gate, `tsgo`, all 19 snapshot tests, and targeted Biome checks passed. Verified against freshly fetched `pascalorg/editor` main: the merge-base is `bcdf422a23580bf533893a73c2aba2192b05a762`. The entire `ViewerSceneContent` component matches that baseline byte-for-byte; the remaining `index.tsx` diff only changes preset isolation selection and transparency.

---

> [!NOTE]
> **Medium Risk**
> Touches core WebGPU snapshot rendering, per-frame isolation/fog handling, and a pinned third-party Environment plugin—high user-visible impact but scoped to capture and presentation paths with added tests.
> 
> **Overview**
> **Snapshot and preset capture** now match what users see in the viewport: offscreen pipelines take an optional `useSceneAtmosphere()` source so thumbnails use the active presentation sky (native or Environment) instead of always compositing the scene-theme gradient. `createSnapshotPipeline` adds a custom scene pass that forces transparent clears inside Three.js r186 nested RTTs, fixing opaque skies on preset captures.
> 
> **Preset isolation** is reworked around `refreshIsolation`, which reconciles layer masks on the full live scene (including late-mounted presentation geometry and instance batches), strips fog for the isolated draw without touching reflections, and restores it after each frame. The editor turns on viewer transparency only while preset isolation is active. Thumbnail capture calls `refreshIsolation` immediately before rendering so new meshes respect the filter.
> 
> **Environment** is pinned to commit `1aa0bc3` (global master control, isolation-aware visibility, Three.js 0.186). Changelog entries document the snapshot fix and plugin bump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 39cc8d4b1ebe862bb1721745fb0d7e949f9e4bf9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AxiomeCG:
**Isolation completeness added** (`39cc8d4b`): `applyIsolation` ran once over `sceneRegistry` roots, so geometry mounted afterwards — Environment presentations, sky, ground, hosted surfaces, instance batches — stayed on the scene layer and leaked into isolated captures.

Now `refreshIsolation(scene)` re-applies the keep-set immediately before every render path (viewport, fallback, snapshot), tracks held objects for idempotent restore, suppresses `fog`/`fogNode` for the synchronous draw (restored after, never held across async readback), and keeps isolated objects out of the shadow-only pass.

11 new isolation tests (`isolation.test.ts`, all green, 350/350 viewer suite), editor `tsgo` clean, workspace build green. The worktree used for manual QA is synced to this commit.

PR #888, opened 2026-09-17
editor: keep GLB and USDZ exports from hanging

Aymericr:
## What does this PR do?

Fixes the Export panel getting stuck on "Exporting GLB…" / "Preparing GLB file…" with nothing downloading (reported by a user of the open-source editor on a two-level, furnished scene). The panel's `try/catch/finally` is sound, so the only way to reach that state is an export promise that never settles, and the portable export path had two of those:

- **Per-texture GPU renderer.** Both the GLB and USDZ exporters used three's `WebGPUTextureUtils.decompress` for every compressed KTX2 texture. It constructs, initialises (adapter + device request, **no timeout**) and disposes a whole `WebGPURenderer` per texture; a furnished scene pays that dozens of times and a single stalled device request parks the export silently. Passing it a renderer is not an option: it resizes the renderer it is given (the live viewer's canvas) and, in r186, throws on the first call that supplies one (`_renderer` starts `undefined`, the cleanup checks `!== null`).
- **Unreportable exporter failures.** `GLTFExporter`'s binary path completes inside `FileReader.onloadend` callbacks with no error handler, after `writeAsync` has already resolved, and wraps `canvas.toBlob` in a promise that never rejects. A failed read calls neither `onDone` nor `onError`; the final read can hand `onDone` `null`, which `new Blob([null])` turned into a four-byte download.

Changes:

- `export-texture-utils.ts` (new): one bounded GPU decompressor per export, shared by preparation (normal-map baking) and serialisation, for GLB and USDZ. Built on the viewer's `initializeGpuRenderer`, so it gets the same adapter/init timeouts and WebGL fallback as the viewer, releases its device on dispose, and refuses work after dispose.
- `glb-export.ts`: every export runs under `withExportDeadline` (3 min default, `timeoutMs` option, `Infinity` disables). Covers plugin `bakeGeometryAsync` / presentation builders as well as the exporter. A `null` result rejects instead of resolving. `serializePreparedSceneToGlb` takes the shared `textureUtils`.
- `usdz-export.ts`: same decompressor and deadline.
- `bake-exporter.tsx`: passes `timeoutMs: Infinity`; the bake worker owns its deadline (`BAKE_CAPTURE_TIMEOUT_MS`), so headless bake behaviour is unchanged.
- Tests: `export-texture-utils.test.ts` (single renderer across textures, clamping, start-up failure, post-dispose guard) and `glb-export-serialize.test.ts` (drives the real `GLTFExporter` with a DOM stand-in: FileReader failure → deadline rejection, final-read `null` → "produced no data", compressed textures routed through the shared decompressor, never-settling plugin hook → deadline, preparation and serialisation share one decompressor).

Follow-ups deliberately left out (tracked privately): upstream three fix for `FileReader.onerror` in `GLTFExporter`, capping `maxTextureSize` / JPEG for opaque maps to shrink portable GLBs, `downloadBlob` revoke timing.

Refs #831 (introduced the portable export path), #867 (added the second `WebGPUTextureUtils` caller).

## How to test

1. `cd packages/editor && bun test src/lib/glb-export-serialize.test.ts src/lib/export-texture-utils.test.ts` — 14 tests, all exercising the real `GLTFExporter` / decompressor.
2. `bun test src` in `packages/editor` (1050 pass) and `bun run check-types` (clean).
3. In the editor with a scene using KTX2 finishes and furniture: Settings → Export GLB and Export USDZ. Expected: files download, button returns to "Export GLB", no error line. Textures in the GLB match the viewer (verified in headless Chromium against `public/material/leather/calf_leather/*.ktx2`: pixel-identical to the stock decompressor).
4. Regression guard for the hang: temporarily make a plugin's `bakeGeometryAsync` return a never-settling promise, export, wait 3 min → the panel shows "GLB export timed out after 180 s…" and the buttons re-enable.

Timing for eight decompress calls in headless Chromium, shared renderer vs stock per-texture renderer:

| Backend | Shared | Stock |
|---|---|---|
| SwiftShader | 18 ms | 359 ms |
| Metal | 40 ms | 501 ms |

N/A — non-visual change (export pipeline). The user's report screenshot shows the stuck state: "Exporting GLB…" button with "Preparing GLB file…" status and the sibling export buttons disabled.

## Checklist

- [ ] I've tested this locally with `bun dev` (verified via unit tests and a headless-Chromium harness; the standalone app was not runnable from the monorepo checkout used here)
- [x] I've updated relevant documentation (if applicable) — none needed; behaviour documented in code comments

---

> [!NOTE]
> **Medium Risk**
> Touches the GLB/USDZ export pipeline and GPU lifecycle during export; incorrect deadline or decompressor disposal could fail exports or leak GPU resources, but behavior is covered by new tests and defaults preserve interactive timeout while bake uses Infinity.
> 
> **Overview**
> Fixes the export UI staying on “Exporting…” when the underlying promise never settles. **GLB and USDZ** now use a single **`createExportTextureUtils`** decompressor per export (shared by normal-map prep and `GLTFExporter`/`USDZExporter`) instead of three’s per-texture **`WebGPUTextureUtils`**, which repeatedly spun up WebGPU and could stall indefinitely.
> 
> Exports run under **`withExportDeadline`** (default **180 s**, overridable via **`timeoutMs`**; **`Infinity`** disables). Silent **`FileReader`** / exporter failures are handled by rejecting on a **null** GLB buffer and by timing out stuck preparation (e.g. never-settling **`bakeGeometryAsync`**). **`BakeExporter`** passes **`timeoutMs: Infinity`** so the bake worker keeps its own deadline.
> 
> Adds unit tests for the decompressor and GLB serialization path; bumps one core temporal-reconciliation test to **30 s** for CI. **`ExportTextureUtils`** is exported from the editor package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d44578b06423edf852d3ef2ee128830a62aa118. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


Aymericr:
CI note: the first `quality` run failed on `packages/core/src/store/use-scene-temporal-reconciliation.test.ts` ("Maxi 8× endpoint undo/redo…", 5.7 s against bun's 5 s default timeout). The last three `main` runs fail on the same test with the same timeout, so it is a pre-existing runner flake unrelated to this change; the rerun passed.

Aymericr:
Added two small commits on top: `refactor(editor)` makes `exportSceneToGlb`/`exportSceneToUsdz` the single deadline owner and exports the `ExportTextureUtils` type; `test(core)` gives the 8× undo/redo fixture a 30 s timeout since it runs 5–6 s on the shared runner and has been failing every `main` CI run since 2026-09-16 (including two of this PR's three runs).

